PR #111, opened 2014-10-09
Add XML Support

kraman:
/to @ritch


lib/http-context.js
```diff
@@ -269,6 +277,39 @@ HttpContext.prototype.done = function () {
...
+        } else {
+          res.header('Content-Type', 'text/xml');
+        }
+        if (data === null) {
+          res.header('Content-Length', '7');
+          res.end('<null/>');
+        } else {
+          try {
+            var input = data;
+            if (Object.prototype.toString.call(input) === '[object Array]') {
```
raymondfeng: See http://web.mit.edu/jwalden/www/isArray.html

PR #148, opened 2014-12-10
Ensure Remote-Connector Converts Types in Array

BerkeleyTrue:
Fixes strongloop/loopback#886


ritch:
LGTM

Thanks for the patch!

PR #153, opened 2015-01-11
Fix errant console.log

STRML:
Not sure how this made it through review, please be careful about this!


STRML:
Going to close this in favor of #154.

PR #155, opened 2015-01-11
Fix parsing JSON arrays with arrayItemDelimiters on.

### LINKED ISSUE #154: Using `arrayItemDelimiters` properly
Is there a good way to use `arrayItemDelimiters` and maintain support for accepting JSON arrays?

I want to be able to accept `["a","b","c"]`, as well as `a,b,c` in a querystring.

If not, I'll take a harder look at #142 and put together a patch.

STRML:
Fixes #154 and obviates #153.


bajtos:
LGTM. @ritch could you PTAL too, to make sure I did not miss any edge case?

PR #156, opened 2015-01-13
Optimize the code to remove closure and bind

raymondfeng:
/to @bajtos or @ritch 
/cc @bnoordhuis

### DIFF
--- a/lib/http-context.js
+++ b/lib/http-context.js
@@ -61,11 +61,10 @@ HttpContext.prototype.buildArgs = function(method) {
   var args = {};
   var ctx = this;
   var accepts = method.accepts;
-  var returns = method.returns;
-  var errors = method.errors;
 
   // build arguments from req and method options
-  accepts.forEach(function(o) {
+  for (var i = 0, n = accepts.length; i < n; i++) {
+    var o = accepts[i];
     var httpFormat = o.http;
     var name = o.name || o.arg;
     var val;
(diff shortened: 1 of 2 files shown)


ritch:
Did this result in any serious performance improvements?

The changes seem fine to me.

raymondfeng:
I tested the changes with JMeter on 16 threads. Throughput increased from 2383/s to 2576/s.

lib/http-context.js
```diff
@@ -192,27 +191,45 @@ function coerce(str) {
...
+  var i;
+  var n;
 
   switch (type) {
     case 'string':
       return coerce(obj);
     case 'object':
       if (obj) {
-        Object.keys(obj).forEach(function(key) {
+        var props = Object.keys(obj);
```
bnoordhuis: If obj is prototype-less or if you don't care that prototype properties get iterated over, using `for (var key in obj)` is usually faster.

lib/http-context.js
```diff
@@ -192,27 +191,45 @@ function coerce(str) {
...
+        for (i = 0, n = props.length; i < n; i++) {
```
bnoordhuis: I don't know what the house style is but just you can just use `var i` here and below.  The declaration gets hoisted to the top and it doesn't matter from a performance perspective.

lib/http-context.js
```diff
@@ -192,27 +191,45 @@ function coerce(str) {
...
+        coerceAll(obj[i]);
+      }
       break;
   }
 
   return obj;
 }
 
+function buildArgs(ctx, method, fn) {
+  var args = ctx.args || {};
```
bnoordhuis: This assignment looks dead to me.  It's reassigned in the try block and it's unused in the catch block.

raymondfeng: Good catch.

lib/http-context.js
raymondfeng: We declare the variables at the top.

lib/http-context.js
raymondfeng: I was following https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#5-for-in

bnoordhuis: Petka is not wrong but it's a tradeoff.  Garbage collection is frequently the biggest performance killer, hence my advice.  Try tracing it with --trace_opt and --trace_deopt.  If it does deoptimize, move it into a separate function.

PR #157, opened 2015-01-13
Optimize performance of http-context

bajtos:
Replace `.forEach()` + `.bind()` with a regular for loop.

/to @raymondfeng please review

Note: the problem addressed by this patch was pointed out in strongloop/loopback-benchmarks#1 as a possible improvement. I run the benchmark again with the patched version of loopback, but I did not observe any performance improvement greater that the variations in the measured data.


bajtos:
Closing in favour of #156.

PR #163, opened 2015-01-18
Add proper support for remotes using HTTP authentication

rmg:
* [x] Add tests for existing Basic auth support (to formalize support and guard against regressions)
 * [x] Add tests for Digest auth
 * [x] Add support for Digest auth remotes

This is necessary to allow loopback-connector-remote to support backends that use Digest auth, which is required to allow strong-pm (or any other backend) to support Digest auth.

/cc @sam-github @kraman

### DIFF
--- a/test/auth.test.js
+++ b/test/auth.test.js
@@ -1,0 +1,96 @@
+var auth = require('http-auth');
+var crypto = require('crypto');
+var expect = require('chai').expect;
+var express = require('express');
+var fmt = require('util').format;
+
+var RemoteObjects = require('../');
+var User = require('./e2e/fixtures/user');
+
+describe('support for HTTP Authentication', function() {
+  var server;
+  var remotes = RemoteObjects.create();
+  remotes.exports.User = User;
+
+  before(function setupServer(done) {
+    var app = express();
+    var basic = auth.basic({ realm: 'testing' }, function(u, p, cb) {
+      cb(u === 'basicuser' && p === 'basicpass');
+    });
+    var digest = auth.digest({ realm: 'testing' }, function(user, cb) {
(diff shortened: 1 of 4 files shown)


rmg:
@bajtos not sure who to assign to, but I figure you'll be online before anyone else. If you aren't the best to review, can you reassign for me if you know who is best to review?

bajtos:
I'd say @ritch is the best person to review this change, I am reassigning to him.

/cc @raymondfeng

test/auth.test.js
```diff
@@ -0,0 +1,96 @@
+var auth = require('http-auth');
```
ritch: Does this work in the browser?

ritch: Not necessarily required to land this, just curious.

ritch:
The changes look good. We'll need to document this (@crandmck) here and in the remote connector.

:+1:

test/auth.test.js
rmg: Not sure. I didn't look into it because I assumed the tests were node-only, and http-auth is only used for the tests.

rmg:
@crandmck for documentation, the PR can be summarized as:
 * strong-remoting will honour the `user:pass` portion of `http://user:pass@resthost/api`
 * supports both Basic auth and Digest auth schemes
 * loopback-connector-remote will inherit this capability once strong-remoting is published

For background, all we are doing is setting `sendImmediately: false` so that [Digest auth works as documented in request](https://github.com/request/request#http-authentication) automatically.

crandmck:
Added https://github.com/strongloop/loopback-connector-remote/issues/13 for docs.   At this point, my focus is primarily on how this is used in LoopBack.

PR #162, opened 2015-01-15
Remove use of deprecated req.param()

aars:
Express 4.11.0 adds deprecation warnings about this usage. Refactor
HttpContext.getArgByName() to access req.params, req.body and
req.query directly.

PR #164, opened 2015-01-19
Update use of request.param(name) to request.params[name]

bradplank:
Express deprecated the use of request.param(name).

Signed-off-by: bradplank <brad@bradplank.com>


slnode:

bajtos:
Hi @bradplank, thank you for the contribution. I am afraid your fix is changing the behaviour, `app.param('foo')` is not the same as `app.params[foo]` AFAIK.

I am closing this pull request in favour of #162.

bradplank:
Thanks, that makes sense to me.

PR #166, opened 2015-01-28
fix express deprecated req.param(name)

paulomcnally:
express deprecated req.param(name): Use req.params, req.body, or req.query instead node_modules/loopback/node_modules/strong-remoting/lib/http-context.js:142:18
    express deprecated req.param(name): Use req.params, req.body, or req.query instead node_modules/loopback/node_modules/strong-remoting/lib/http-context.js:153:22

### DIFF
--- a/lib/http-context.js
+++ b/lib/http-context.js
@@ -139,7 +139,7 @@ HttpContext.prototype.buildArgs = function(method) {
 
 HttpContext.prototype.getArgByName = function(name, options) {
   var req = this.req;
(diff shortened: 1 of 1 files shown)


paulomcnally:
Duplicate :/ https://github.com/strongloop/strong-remoting/pull/162

bajtos:
Closing in favour of #162. Note that this patch is a breaking change, because AFAIK `req.param(name)` is not the same as `req.params(name)`.

Issue #167, opened 2015-01-30
Better argument coercion


ritch:
See https://github.com/strongloop/strong-remoting/issues/143

~~I'd like to add support for explicit argument coercion using the type defined in accepts.~~

**Edit: we already support this via `Dynamic`.**

For non-typed arguments (eg. the `id` property in the url path of `PersitedModel.create()`)... we should support intelligent coercion. This should follow this basic algorithm:

1. if we know the type, coerce using `Dynamic`
2. if the type is undefined or `any` determine the type using regex / comparison
3. if an integer pattern is matched - ensure the number does not start with 0

/cc @BerkeleyTrue

Issue #143, opened 2014-12-06
coerce id "0668" detects integer, returns 688;


BerkeleyTrue:
Using model id's that allow for leading zero's and all number characters causes the coerce function in http-context.js [here](https://github.com/strongloop/strong-remoting/blob/67d4b139f323ce2a5e7a2e785320f66776275e92/lib/http-context.js#L188) to change model ids to integers and remove leading zeros.

This was causing a new model to be created with id 688 and model '0688' was not being updated. 

Not sure if this was intended, but I will be making sure all id's require at least one alpha character to avoid this, but wanted to have some record of this for future reference.

ritch:
Here is what I'd like to do to fix this:

1. Take the type of the argument into account to allow explicit type coercion
2. For non-typed arguments... support intelligent coercion using @BerkeleyTrue's suggestion (leading with a 0 is a string)

I would consider 1 an enhancement and 2 a bug fix. I'll create separate issues to track these changes.

richardpringle:
Looks like this issue has been resolved. Please re-open if there are any further problems. 

@bajtos, I know that you have the story to deal with some coercion issues in strong-remoting but I believe that this is not related. Please correct me if I am wrong.

bajtos:
The issue is fixed in the upcoming 3.0 release.

- When the argument's type is `any`, strong-remoting will treat integer numbers starting with `0` as string, e.g. `0668` will preserve the leading zero. However, floats are still converted to numbers, e.g. both `0.42` and `00.42` are converted to a number.

- When the argument's type is `number` or `integer`, strong-remoting will treat integer numbers starting with `0` as numbers and discarding any extra leading zeroes.

bajtos:
Actually it looks like this issue was already fixed in 2.13.0 by #168.

PR #168, opened 2015-01-30
Improve coercion for untyped args

ritch:
See https://github.com/strongloop/strong-remoting/issues/143 and https://github.com/strongloop/strong-remoting/issues/167

/to @raymondfeng @bajtos 
/cc @BerkeleyTrue

### DIFF
--- a/lib/http-context.js
+++ b/lib/http-context.js
@@ -117,6 +117,12 @@ HttpContext.prototype.buildArgs = function(method) {
     // cast booleans and numbers
     var dynamic;
     var otype = (typeof o.type === 'string') && o.type.toLowerCase();
+    var isAny = !otype || otype === 'any';
+
+    // try to coerce dynamic args when input is a string
+    if(isAny && typeof val === 'string') {
+      val = coerce(val);
+    }
 
     if (Dynamic.canConvert(otype)) {
       dynamic = new Dynamic(val, ctx);
@@ -189,7 +195,7 @@ function coerce(str) {
   if ('true' === str) return true;
   if ('false' === str) return false;
   if (isfloat.test(str)) return parseFloat(str, 10);
-  if (isint.test(str)) return parseInt(str, 10);
+  if (isint.test(str) && str.charAt(0) !== '0') return parseInt(str, 10);
   return str;
 }
 
--- a/lib/http-context.js
+++ b/lib/http-context.js
@@ -160,7 +160,7 @@ HttpContext.prototype.getArgByName = function(name, options) {
 
   var arg = (args && args[name] !== undefined) ? args[name] :
             this.req.params[name] !== undefined ? this.req.params[name] :
-            this.req.body[name] !== undefined ? this.req.body[name] :
+            (this.req.body && this.req.body[name]) !== undefined ? this.req.body[name] :
             this.req.query[name] !== undefined ? this.req.query[name] :
             this.req.get(name);
   // search these in order by name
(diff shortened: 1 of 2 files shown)


ritch:
@raymondfeng @bajtos do you have any thoughts here: https://github.com/strongloop/strong-remoting/pull/168/files#diff-d54d77feb22761cf4ef39576b7edd949R29

lib/http-context.js
```diff
@@ -184,7 +190,7 @@ function coerce(str) {
   if ('true' === str) return true;
   if ('false' === str) return false;
   if (isfloat.test(str)) return parseFloat(str, 10);
-  if (isint.test(str)) return parseInt(str, 10);
+  if (isint.test(str) && str.charAt(0) !== '0') return parseInt(str, 10);
```
bajtos: IIUC, this change is not covered by unit-tests, as I don't see any test for an int-value starting with a zero, e.g. "0123".

test/http-context.test.js
```diff
@@ -0,0 +1,89 @@
...
+        expectedValue: 0.123456
+      }));
+
+      // NEEDS DISCUSSION: this test fails...
+      // not sure what we should do here - add an option to accept "null" as `null`?
+      // it('should coerce null strings into null', givenMethodExpectArg({
+      //   type: 'string',
+      //   input: 'null',
+      //   expectedValue: null
+      // }));
```
bajtos: Well, since we cannot really tell whether the user wanted to pass a string with the value `"null"`, or a null string, I'd say let's stick to the current behaviour.

You should uncomment the test and fix the expected value to describe the current behaviour. That way we won't accidentally introduce a regression in the future.

Please consider adding `null` tests for other types like `int` and `float` too, so that we have a complete running specification of strong-remoting's behaviour.

test/http-context.test.js
```diff
@@ -0,0 +1,89 @@
...
+      }
+      done();
+    });
+
+    request(app).get('/?testArg=' + options.input).end();
+  }
+}
+
+
+function createMethod(type) {
```
bajtos: The name is confusing, what kind of method is this creating? `createMethodWithAnArgumentOfType` would reveal the intent better.

test/http-context.test.js
```diff
@@ -0,0 +1,89 @@
...
+      var ctx = new HttpContext(req, res, method);
+      try {
+        expect(ctx.args.testArg).to.equal(options.expectedValue);
+      } catch(e) {
+        return done(e);
+      }
+      done();
+    });
+
+    request(app).get('/?testArg=' + options.input).end();
```
bajtos: `testArg` is coupled with implementation details in `createMethod`, it's not clear where is this magic string coming from.

Considering my other comment on L83, I'd say you should inline `createMethod` as it does not really improved the code.

lib/http-context.js
ritch: Added a test for this... Thanks

test/http-context.test.js
ritch: Removed...

test/http-context.test.js
ritch: Removed

ritch:
> If an arg is untyped or typed as any, do we still want to coerce the value?

Only when it is obvious what to coerce it to. In cases where coercion is not doing what you want... you can just provide the correct type for the arg since you know what it should be.

ritch:
@bajtos @raymondfeng ping... this should be ready to merge

bajtos:
The code LGTM. However, Jenkins CI fails on all slaves. Please investigate and fix before landing this patch.

ritch:
test please

ritch:
test please

PR #169, opened 2015-01-30
Update simple example

ritch:
The examples are out of date. This corrects the simple example. The others should take the same approach.

Although before that I'd like to add some sugar APIs to make the examples a bit less redundant.

/to @raymondfeng @bajtos 
/cc @jakerella


jakerella:
This works for me, gracias!

:shipit:

raymondfeng:
LGTM

PR #172, opened 2015-02-02
Catch unhandled error on current method invocation

PradnyaBaviskar:
to/ @bajtos 
Made following changes for issue#709
- In remote-objects.js Enclosed the call to cur.call with try-catch block
- Added test case to verify that unhandled exceptions, do not crash the server.

Can you please review?

### DIFF
--- a/lib/remote-objects.js
+++ b/lib/remote-objects.js
@@ -359,7 +359,11 @@ RemoteObjects.prototype.execHooks = function(when, method, scope, ctx, next) {
     var cur = stack.shift();
 
     if (cur) {
-      cur.call(scope, ctx, execStack, method);
+      try {
+        cur.call(scope, ctx, execStack, method);
+      } catch (err) {
+        next(err);
+      }
     } else {
       next();
     }
--- a/test/rest.test.js
+++ b/test/rest.test.js
@@ -1184,6 +1184,23 @@ describe('strong-remoting-rest', function() {
           .expect(500)
           .end(expectErrorResponseContaining({message: 'an error'}, done));
       });
+
+      it('should return 500 for unhandled errors thrown from before hooks', function(done) {
+        var method = givenSharedStaticMethod();
+
+        objects.before(method.name, function(ctx, next) {
+          process.nextTick(next);
+        });
+
+        objects.before(method.name, function(ctx, next) {
+          throw new Error('test error');
+        });
+
+        request(app).get(method.url)
+          .set('Accept', 'application/json')
+          .expect(500)
+          .end(expectErrorResponseContaining({ message: 'test error' }, done));
+      });
     });
 
     it('should return 500 when method returns an error', function(done) {


raymondfeng:
LGTM. @bajtos PTAL.

lib/remote-objects.js
```diff
@@ -359,7 +359,9 @@ RemoteObjects.prototype.execHooks = function(when, method, scope, ctx, next) {
     var cur = stack.shift();
 
     if (cur) {
-      cur.call(scope, ctx, execStack, method);
+      try {
+        cur.call(scope, ctx, execStack, method);
+      } catch (err) { next(err); }
```
bajtos: Nitpick: I believe we place the body of `catch` statement on a new line:

```js
} catch (err) {
  next(err);
}
```

test/rest.test.js
```diff
@@ -1184,6 +1184,33 @@ describe('strong-remoting-rest', function() {
           .expect(500)
           .end(expectErrorResponseContaining({message: 'an error'}, done));
       });
+      
+      it('should return 500 for unhandled errors', function(done) {
```
bajtos: Please improve the test name and mention the important part - you are testing unhandled errors in `before` hooks.

test/rest.test.js
```diff
@@ -1184,6 +1184,33 @@ describe('strong-remoting-rest', function() {
...
+        var method = givenSharedStaticMethod(
+          function greet(name, cb) {
+            cb(null, name );
+          },
+          {
+            accepts: { arg: 'name', type: 'string', http: {source: 'body'} },
```
bajtos: Please simplify the test method - remove the unused argument.

test/rest.test.js
```diff
@@ -1184,6 +1184,33 @@ describe('strong-remoting-rest', function() {
...
+        });
+       
+        objects.before(method.name, function(ctx, next) {
+          throw new Error('test error');
+        });       
+       
+        request(app).get(method.url)
+          .set('Accept', 'application/json')
+          .set('Content-Type', 'application/json')
+          .send({name: 'abc'})
```
bajtos: In the same spirit as above, remove `Content-Type` and request body (`.send()`) as they are not necessary.

PradnyaBaviskar:
to / @bajtos,
Made corrections, as suggested in your comments. Can you please review?

test/rest.test.js
```diff
@@ -1184,6 +1184,27 @@ describe('strong-remoting-rest', function() {
           .expect(500)
           .end(expectErrorResponseContaining({message: 'an error'}, done));
       });
+      
+      it('should return 500 for unhandled errors thrown from before hooks', function(done) {
+        var method = givenSharedStaticMethod(
+          function greet(cb) {
+            cb();
+          }
+        );
```
bajtos: Nice, this his much simpler and better.

One more improvement to make: when your shared method is a no-op like in this case, you don't need to provide any function to `givenSharedStaticMethod`, it will build one by default - see [test/rest.test.js#L1975](https://github.com/PradnyaBaviskar/strong-remoting/blob/fd721cfac710f23ff9d88854cba66fa3b08e5764/test/rest.test.js#L1975).

test/rest.test.js
```diff
@@ -1184,6 +1184,27 @@ describe('strong-remoting-rest', function() {
...
+        });
+       
+        objects.before(method.name, function(ctx, next) {
+          throw new Error('test error');
+        });       
+       
+        request(app).get(method.url)
+          .set('Accept', 'application/json')
+          .expect(500)
+          .end(expectErrorResponseContaining({message: 'test error'}, [''], done));
```
bajtos: Coding style: we put spaces inside curly brackets (objects). (But we don't put spaces inside square brackets for array). 

(I do realise the check on the line 1186 is not conforming to this rule, our coding style is unfortunately not very consistent ATM.)

Also the second argument `['']` is IMO not needed.

```
.end(expectErrorResponseContaining({ message: 'test error' }, done));
```

bajtos:
Two more details to fix. Please squash the commits into a single one when you are done, so that I can merge this PR via the GitHub UI.

PradnyaBaviskar:
to / @bajtos ,
Made fixes to two more changes suggested above
- Changed the shared method to givenSharedStaticMethod()
- Added spaces withing curly braces
Please review.

bajtos:
Landed, thank you!

PR #174, opened 2015-02-05
Set 302 status code if response has Location header

umbrella-ivan-zhukov:
for loopback issue strongloop/loopback#1065

### DIFF
--- a/lib/http-context.js
+++ b/lib/http-context.js
@@ -375,10 +375,7 @@ HttpContext.prototype.done = function() {
     if (!res.get('Content-Type')) {
       res.header('Content-Type', 'application/json');
     }
-    // Location header is set?
-    if (res.get('Location') !== undefined) {
-      res.statusCode = 302;
-    } else {
+    if (res.statusCode === 200) {
       res.statusCode = 204;
     }
     res.end();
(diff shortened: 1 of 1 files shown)


slnode:

bajtos:
@umbrella-ivan-zhukov thank you for the pull request. Please add a unit-test to verify that the implementation works as intended and to show how is this intended to be use.

I don't understand one thing: since the Location header was already set, why cannot the code which is changing this header also change the status code?

Perhaps the following implementation would be more correct?

```js
// in the "no content/data" branch
if (res.statusCode === 200) 
  res.statusCode = 204;
```

umbrella-ivan-zhukov:
@bajtos , yes you're right. I'll push update to loopback PR.
But I couldn't find "no content/data" branch. Should I create it?

bajtos:
> But I couldn't find "no content/data" branch. Should I create it?

Let me explain myself more clearly. I meant that my code above should be in the "else" branch on lines 375-379, i.e. just add the if statement before the line `res.statusCode = 204`.

Does it make more sense now?

umbrella-ivan-zhukov:
Updated the code.
I don't have time for test, but I'll add one this weekend.

ritch:
I'm going to close this since the change is only 3 lines and is missing a test. I'll create an issue to track this though. Please open another pull request with the correct title, test, and changes.

ritch:
See https://github.com/strongloop/strong-remoting/issues/175

PR #179, opened 2015-02-13
Support Promise-returning method and hooks

bajtos:
If a shared method or a hook handler returns a duck-typed promise, i.e. an object with a `then` method, then the invocation waits until the promise is fulfilled or rejected.

Connect strongloop/loopback#418

/to @raymondfeng @ritch please review


raymondfeng:
LGTM.

PR #180, opened 2015-02-13
Allow promises to resolve with a single arg

bajtos:
A follow-up for #179, connect strongloop/loopback#418

Comparison of callback and promise APIs:

```
callback(null, value) <===> resolve(value)
callback(null, arg1, arg2) <===> resolve([arg1, arg2])
```

/to @raymondfeng @ritch please review


raymondfeng:
LGTM

Issue #105, opened 2014-09-18
HttpInvocation.invoke() should support authentication


coodoo:
When replicating local model with remote models, it will trigger REST end points on remote server, sometimes it would be necessary to pass TOKEN in the request header for authorization, it's not possible with current implementation.

Ideally, TOKEN should be passed down all the way from persisted-model when invoking, or better yet, token can be stored somewhere (probably app.settings{} or app.configs{}) for easier access.

Here's the pseudo code of how it works if TOKEN were passed down:

In http-invocation.js:
```js
HttpInvocation.prototype.invoke = function (callback) {
  var req = this.createRequest();
  
  // append TOKEN to req.headers, in this case, I storing it in window{} earlier, which of course is a hack
  req.headers = {'Authorization': window.zoken};

  request(req, function(err, res, body) {
    if(err instanceof SyntaxError) {
      if(res.status === 204) err = null;
    }
    if(err) return callback(err);
    this.transformResponse(res, body, callback);
  }.bind(this));
}
```

PS. just reviewed the whole replication/diff/merge implementation, really great work, two thumbs up! Can't wait to use it in my next project.

bajtos:
Related? strongloop/loopback-connector-remote#11

ritch:
`remotes.auth` released as **2.13.0**

bajtos:
Here is a code snippet showing how to set the access token in a loopback application, assuming that the data-source backed by the remote connector is called "remote":

    app.dataSources.remote.connector.remotes.auth = {
      bearer: new Buffer(token).toString('base64'),
      sendImmediately: true
    };

@ritch I find this rather difficult to use, it took me quite a lot of debugging to find out where exactly the token should be set. I think we should add some sugar API to make this easier.

@crandmck Is there any documentation explaining how to configure the remoting connector? I'd like to include this information there.

bajtos:
Let's follow up here: https://github.com/strongloop/loopback-connector-remote/issues/3

crandmck:
@bajtos Here is the doc for the remote connector: http://docs.strongloop.com/display/LB/Remote+connector  Feel free to add something there until https://github.com/strongloop/loopback-connector-remote/issues/3 is resolved.

bajtos:
@crandmck done, I have added a new section named "Configuring authentication". Feel free to improve the content as you wish.

crandmck:
Thanks. LGTM.

PR #181, opened 2015-02-17
Add remotes.auth / authorization support

ritch:
This adds a new feature (built on top of the existing auth support). Using `remotes.auth` you can now provide basic, bearer, and digest authorization (even after `remotes.connect()`).

This patch introduces the api (the `remotes.auth` property) and support in `HttpInvocation`. It maintains backwards compatibility by setting `remotes.auth` using the url provided in `remotes.connect()`.

The key new feature is that authorization can be set at any time and requests afterward will honor this setting. This makes it possible to do this:

```js
remotes.connect('http://localhost:3000');

remotes.invoke('User.login', [{
  username: 'foo',
  password: 'bar'
}], function(err, result) {
  remotes.auth = {
    bearer: result.accessToken
  };
});
```

...all subsequent invocations will be signed with the token.

/to @raymondfeng 
/cc @rmg @bajtos 

Fixes https://github.com/strongloop/strong-remoting/issues/105

### DIFF
--- a/lib/http-invocation.js
+++ b/lib/http-invocation.js
@@ -37,23 +37,9 @@ var JSON_TYPES = ['boolean', 'string', 'object', 'number'];
  * @property {Array} args The arguments to be used when invoking the `SharedMethod`
  */
 
-function HttpInvocation(method, ctorArgs, args, base) {
+function HttpInvocation(method, ctorArgs, args, base, auth) {
   this.base = base;
-  if (this.base) {
-    var url = urlUtil.parse(base);
-    // If base url contains auth, extract it so we can set it separately
-    if (url.auth) {
-      this.username = url.auth.split(':')[0];
-      this.password = url.auth.split(':')[1];
-      // set base without auth so request honours our auth options
-      delete url.auth;
-      this.base = urlUtil.format(url);
-      // ensure a "/" hasn't been appended where there wasn't one before
-      if (base[base.length - 1] !== this.base[this.base.length - 1]) {
-        this.base = this.base.slice(0, -1);
-      }
-    }
-  }
+  this.auth = auth;
   this.method = method;
   this.args = args || [];
   this.ctorArgs = ctorArgs || [];
@@ -177,20 +163,23 @@ HttpInvocation.prototype.createRequest = function() {
   var ctorAccepts = null;
   var query;
   var i;
+  var auth = this.auth;
 
   // initial url is the format
   req.url = this.base + method.getFullPath();
 
   // add auth if it is set
-  if (this.username && this.password) {
-    req.auth = {
-      // request defaults to sending auth immidiately, which works for
-      // Basic auth, but doesn't work for Digest auth
-      // Note: these options are ignored if the URL still contains credentials
-      sendImmediately: false,
-      user: this.username,
-      pass: this.password,
-    };
+  if (auth) {
+    req.auth = {};
+    if(auth.username && auth.password) {
+      req.auth.username = auth.username;
+      req.auth.password = auth.password;
+    }
+    if('sendImmediately' in auth) {
+      req.auth.sendImmediately = auth.sendImmediately;
+    } else {
+      req.auth.sendImmediately = false;
+    }
   }
 
   // build request args and method options
--- a/test/auth.test.js
+++ b/test/auth.test.js
@@ -55,6 +55,15 @@ describe('support for HTTP Authentication', function() {
        fails('/digestAuth'));
   });
 
+  describe('remotes.auth', function () {
+    it('should be populated from the url', function () {
+      var url = 'http://login:pass@myhost.com'
+      remotes.connect(url, 'rest');
+      expect(remotes.auth.username).to.eql('login');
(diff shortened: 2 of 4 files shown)


lib/http-invocation.js
```diff
@@ -177,20 +163,23 @@ HttpInvocation.prototype.createRequest = function() {
...
-      user: this.username,
-      pass: this.password,
-    };
+  if (auth) {
+    req.auth = {};
+    if(auth.username && auth.password) {
+      req.auth.username = auth.username;
+      req.auth.password = auth.password;
+    }
+    if('sendImmediately' in auth) {
```
rmg: This appears to be a boolean with a default of `false`. Would it be clearer to do this instead?
```js
req.auth.sendImmediately = !! auth.sendImmediately;
```

When I see an explicit `if`/`else` like the one here it makes me shift gears and start looking for what makes it not just a simple boolean. That could just be me, though.

rmg:
Your example in the PR description shows setting credentials for all future requests. Is it also possible to override the auth on a per-request basis? I'm imagining a `sudo`-like scenario where one wants to use unprivileged access for "read" operations and only send credentials for "write" operations.

ritch:
> Your example in the PR description shows setting credentials for all future requests. Is it also possible to override the auth on a per-request basis? I'm imagining a sudo-like scenario where one wants to use unprivileged access for "read" operations and only send credentials for "write" operations.

Not possible, but we can add it. It would probably look something like this:

```js
remotes.invoke('User.greet', [], {bearer: '...'}, cb);
```

lib/http-invocation.js
bajtos: +1

test/auth.test.js
```diff
@@ -55,6 +55,16 @@ describe('support for HTTP Authentication', function() {
        fails('/digestAuth'));
   });
 
+  describe('remotes.auth', function () {
+    it('should be populated from the url', function () {
+      var credentials = 'foo:bar';
+      var url = fmt('http://%s@myhost.com', credentials);
```
bajtos: Is there any particular reason for using `fmt` instead of including the credentials in the url string? Also `foo:bar` is not very descriptive,

Proposal:

```js
var url = 'http://login:pass@myhost.com';
```

test/auth.test.js
```diff
@@ -94,3 +110,21 @@ function md5(str) {
...
+    if (credentials.length > 0) {
+      credentials += '@';
+    }
+    var url = fmt('http://%s127.0.0.1:%d%s', credentials, port, path);
+    var method = 'User.login';
+    var args = [{username: 'joe', password: 'secret'}];
+    remotes.connect(url, 'rest');
+    remotes.invoke(method, args, callback);
+  }
+*/
```
bajtos: Please remove the commented-out code.

bajtos:
The Android and AngularJS SDKs update the auth object automatically on login and logout. I am wondering if this can be achieved in the isomorphic case with the solution proposed in this PR? Can we add this feature either as part of this PR, or in a follow-up PR?

My guess on what it can look like:

```js
remotes.after('*.login', function(ctx, token, next) {
  // a duck-type check whether token is an AccessToken
  if (!(token && token.id && token.ttl && token.userId)) return; 
  remotes.auth = { bearer: token.id }
});

remotes.after('*.logout', function(ctx, token, next) {
  remotes.auth = null;
});
```

lib/http-invocation.js
ritch: > what makes it not just a simple boolean

This value is passed to `request`. I don't want to coerce any other value into a boolean. It probably will always be a boolean, but I don't see any value in coercion.

ritch:
> My guess on what it can look like:

That looks about right. I will create a PR in `loopback-connector-remote`.

test/auth.test.js
ritch: Agreed. Fixed.

lib/http-invocation.js
bajtos: I find it weird that one branch passes potentially any value type, and the other passes a boolean. Should the else branch set `req.auth.sendImmediately = undefined` or `null` then?

bajtos:
I am still not very happy with the `sendImmediately` part, but it's a detail I can live with.

LGTM.

lib/http-invocation.js
rmg: The complication here is that in request, `sendImmediately` actually defaults to `true`, which breaks Digest Auth (you literally can't know what to send until the server tells you).

simoami:
@ritch It would be great to add tests to illustrate how subsequent calls reuse the already generated token.

I also fail to understand how the token is retained per user if the invocation is triggered across multiple requests.

ritch:
> I also fail to understand how the token is retained per user if the invocation is triggered across multiple requests.

Every time a method is invoked from a client, the `RemoteObjects` (aka `remotes`) `auth` property is checked. If it includes a token, it is added to the appropriate header of the request object.

Setting `remotes.auth` to `null` will clear the token for new requests.

I'm not sure what you mean by "per user". If you would like use a new token for each invocation, you should create a `remotes` for each token.

Issue #175, opened 2015-02-06
Preserve non-200 status when no content will be sent


ritch:
Here is a potential implementation...

https://github.com/strongloop/strong-remoting/pull/174/files

richardpringle:
Looks like this has been solved. Closing

PR #183, opened 2015-02-23
Preserve non-200 status when no content will be sent

PradnyaBaviskar:
to / @bajtos -
In reference to
1) Strong-remoting issue - https://github.com/strongloop/strong-remoting/issues/175 , and
2) LB issue - https://github.com/strongloop/loopback/issues/416

I've made changes to retain non-200 status, when no content is sent.

Can you please review?

### DIFF
--- a/test/rest.test.js
+++ b/test/rest.test.js
@@ -722,6 +722,25 @@ describe('strong-remoting-rest', function() {
         .expect(204, done);
     });
 
+    it('should preserve non-200 status when responding with no content', function(done) {
+       var method = givenSharedStaticMethod(
+         function(ctx, cb) {
+           ctx.res.status(302);
+           cb();
+         },
+        {
+          accepts: [
+            { arg: 'ctx', type: 'object', http: {source: 'context' } }
+          ],
+        }
+       );
+
+      request(app).get(method.url)
+        .set('Accept', 'application/json')
+        .expect(302, done);
+
+    });
+
     it('should accept custom content-type header if respond with 204', function(done) {
       var method = givenSharedStaticMethod();
       objects.before(method.name, function(ctx, next) {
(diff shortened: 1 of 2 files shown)


bajtos:
@PradnyaBaviskar please add a unit-test to verify the change works as intended.

bajtos:
This allows developers to write a remote method that calls the final callback and at the same time sets the response to 304 redirect, either via accepting the `req` object as an argument or via a remote hook.

```js
UserModel.afterRemote('confirm', function(ctx, inst, next) {
  if (ctx.req.query.redirectUrl !== undefined) {
    ctx.res.location(redirectUrl);
    ctx.res.status(302);
  }
  next();
});
```

@ritch you asked me to let you review all strong-remoting changes, could you please take a look?

ritch:
I would also add if status code is not defined set it to 204.

Feel free to merge as is though.

PradnyaBaviskar:
@bajtos ,
1) I've added test case to verify that non-200 status is preserved.
2) Also modified the code to set response status to 200, if it is undefined.

Can you please review?

test/rest.test.js
```diff
@@ -722,6 +722,25 @@ describe('strong-remoting-rest', function() {
...
 
+    it('should preserve non-200 status when responding with no content', function(done) {
+       var method = givenSharedStaticMethod(
+         function(ctx, cb) {
+           ctx.res.status(302);
+           cb();
+         },
+        {
+          accepts: [
+            { arg: 'ctx', type: 'object', http: {source: 'context' } }
```
bajtos: FYI: you can also use `{arg: 'res', type: 'object', http: { source: 'res' } }` and then access the response directly.

bajtos:
Landed, thank you!

PR #184, opened 2015-03-06
Coerce empty string into empty array with arrayItemDelimiters.

STRML:
Before this patch, a hit like `GET /endpoint?fooArray=` would, if `arrayItemDelimiters` were used, resolve `fooArray` with the value `['']`.

This is undesirable - I see two options for what an empty value should serialize to for an array arg:

1. `undefined`,
2. `[]`.

I went with `2` because I think `?key=` has a different meaning than `?`, since you might really want to pass the empty array instead of undefined.

### DIFF
--- a/lib/shared-method.js
+++ b/lib/shared-method.js
@@ -199,7 +199,14 @@ SharedMethod.prototype.invoke = function(scope, args, remotingOptions, cb) {
               delims = new RegExp(delims.map(escapeRegex).join('|'), 'g');
               remotingOptions.arrayItemDelimiters = delims;
             }
-            uarg = uarg.split(delims);
+
+            // If we received the empty string (e.g. &arg=), make it an empty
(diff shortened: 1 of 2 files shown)


bajtos:
LGTM.

@STRML could you please sign the CLA? I can't land the patch without that.

https://cla.strongloop.com/agreements/strongloop/strong-remoting

STRML:
Odd, this is far from my first contribution to strong-remoting. Perhaps it lost my CLA sig?

bajtos:
@STRML thank you.

> Odd, this is far from my first contribution to strong-remoting. Perhaps it lost my CLA sig?

@rmg could you PTAL?

rmg:
@STRML your first contributions were before the CLA was introduced, but we have also been having problems with the CLA app not updating pull requests. I'll be looking in to it.

PR #185, opened 2015-03-06
Fix JSCS errors

superkhau:
@ritch Merging without review as these are simple fixes.

### DIFF
--- a/test/rest.test.js
+++ b/test/rest.test.js
@@ -723,22 +723,25 @@ describe('strong-remoting-rest', function() {
     });
 
     it('should preserve non-200 status when responding with no content', function(done) {
-       var method = givenSharedStaticMethod(
-         function(ctx, cb) {
-           ctx.res.status(302);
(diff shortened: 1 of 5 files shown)


test/rest.test.js
```diff
@@ -723,22 +723,25 @@ describe('strong-remoting-rest', function() {
...
+        }, {
           accepts: [
-            { arg: 'ctx', type: 'object', http: {source: 'context' } }
+            {
+              arg: 'ctx',
+              type: 'object',
+              http: {
+                source: 'context'
+              }
+            }
```
bajtos: This is pretty bad, we should change JSCS settings to allow the single-line variant. Check LoopBack settings which allow this style (e.g. in [lib/persisted-model.js](https://github.com/strongloop/loopback/blob/4a81d060186ef3bdba0aa6d8f35cbdb939bfa661/lib/persisted-model.js#L607-L618)).

PR #187, opened 2015-03-26
browser: remove packages used by server only

bajtos:
Remove the following packages:
 - cors
 - js2xmlparser

See https://github.com/strongloop/loopback/issues/989

/to @ritch please review
/cc @BerkeleyTrue


ritch:
:+1:

PR #188, opened 2015-03-30
Do not require instanceof check for type assertion

ritch:
/to @bajtos 
/cc @BerkeleyTrue


BerkeleyTrue:
:+1: Genius!

bajtos:
LGTM.

I am a bit worried that this may actually hide problems caused by having multiple SharedClass constructors around, thus making these problems much more difficult to troubleshoot.

PR #191, opened 2015-04-02
Forward invocation error details

bajtos:
Most notably always set `err.statusCode`.

Connect to strongloop/loopback#1166

/to @ritch please review


ritch:
LGTM

PR #194, opened 2015-04-02
Fix style errors and undefined context in rest test.

STRML:
This would be caught with a precommit hook that run npm test.


bajtos:
Ha, I have discovered the same problem recently, the fix has been already landed in fca2fb05f44f.

PR #196, opened 2015-04-03
Fix improper type coercions, especially on array types.

STRML:
This moves the `arrayItemDelimiters` configuration into http-context.js,
as it is an HTTP implementation detail like querystring parsing and is
not generalizable to remote methods exposed in other ways.

This PR also makes argument coercion more robust for `any`, `object`,
and `array` types.

I was having an issue with single array arguments being coerced improperly
to numbers - for example, with accepts type of `['string']`, input like `'123456'`
was being coerced to the number `123456`. It should be coerced to `['123456']`.

### DIFF
--- a/lib/http-context.js
+++ b/lib/http-context.js
@@ -34,9 +34,9 @@ function HttpContext(req, res, method, options) {
   this.req = req;
   this.res = res;
   this.method = method;
+  this.options = options || {};
   this.args = this.buildArgs(method);
   this.methodString = method.stringName;
-  this.options = options || {};
   this.supportedTypes = this.options.supportedTypes || DEFAULT_SUPPORTED_TYPES;
 
   if (this.supportedTypes === DEFAULT_SUPPORTED_TYPES && !this.options.xml) {
@@ -69,6 +69,13 @@ HttpContext.prototype.buildArgs = function(method) {
     var name = o.name || o.arg;
     var val;
 
+    // Support array types, such as ['string']
+    var isArrayType = Array.isArray(o.type);
+    var otype = isArrayType ? o.type[0] : o.type;
+    otype = (typeof otype === 'string') && otype.toLowerCase();
+    var isAny = !otype || otype === 'any';
+
+    // This is an http method keyword, which requires special parsing.
     if (httpFormat) {
       switch (typeof httpFormat) {
         case 'function':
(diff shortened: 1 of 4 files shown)


ritch:
This looks good overall. I'll take a closer look and merge soon if I don't find anything else.

Great job!

STRML:
Anything I can do to help this along? We're running this fork in production & this is a serious bug IMO.

STRML:
FYI the merge error is caused by https://github.com/strongloop/strong-remoting/commit/fca2fb05f44fe124d285fe2db768375e94f6bc49, which seems to be a typoed fix to the lint error.

STRML:
Thanks.

ritch:
released as **2.16.1**

ritch:
> Thanks.

No. Thank you :+1:

XinV:
Hi, guys, we are having issues when passing the array as the input param to the API endpoint, do you cover the array as param in your unit testing? 

The request is:
      utils.authorizedCall("post", url, token)
        .send({
          "partnerships" : [9006],
          "groups" : ["sony usa"]
        })
        .expect(200)
        .end(function(err, res){
          console.log("Error", err);

2015-04-15T19:12:10.642Z - error: /api/orgs/sony/groupmemberships statusCode=400, statusDescription=Bad Request, @context=/contexts/Error.jsonld, @id=/errors/BadRequestError, code=4004000
Error [Error: expected 200 "OK", got 400 "Bad Request"]
Res { error: 
   { name: 'Error',
     message: 'invalid value for argument \'groups\' of type \'Array\': sony usa. Received type was string. Error: Unexpected token s',
     '@context': '/contexts/Error.jsonld',
     '@id': '/errors/BadRequestError',
     code: 4004000,
     statusDescription: 'Bad Request',
     status: 400,
     statusCode: 400 } }

bajtos:
@XinV I have submitted a patch to fix the bug, see https://github.com/strongloop/strong-remoting/pull/204

XinV:
Thanks @bajtos , for now we force the dependency to v2.15.0. Will upgrade it once your fix is released.

bajtos:
@XinV the bug should be fixed in `strong-remoting@2.16.2`, please let us know if the new version works for you well.

PR #197, opened 2015-04-08
REST: store model, method and ctorArgs on ctx.req

bajtos:
Modify the http-context (used e.g. by the rest-adapter) to expose additional metadata on the `req` object:

    req.remoteClass = name of the remote class (the LoopBack model)
    req.remoteMethod = name of the method invoked
    req.remoteInstance = arguments for the shared constructor

Examples:

    POST /api/Cars
      remoteClass: Car
      remoteMethod: create
      remoteInstance: { id: undefined }

    PUT /api/Cars/123
      remoteClass: Car
      remoteMethod: prototype.updateOrCreate
      remoteInstance: { id: '123' }

    GET /api/Cars/123
      remoteClass: Car
      remoteMethod: findById
      remoteInstance: { id: '123' }

Connect to strongloop/loopback#1276

In the next step, I'll modify strong-express-metrics to detect these three new request properties and include them in all metrics records.

/to @ritch please review
/cc @altsang @anthonyettinger PTAL too, to make sure this is in line with your requirements.

### DIFF
--- a/lib/http-context.js
+++ b/lib/http-context.js
@@ -45,6 +45,8 @@ function HttpContext(req, res, method, options) {
       return !/\bxml\b/i.test(type);
     });
   }
+
+  this._exposeRequestMeta();
 }
 
 /**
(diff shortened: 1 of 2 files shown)


bajtos:
This does not correctly handle `GET /api/Cars/123`, as `findById` is not recognised as a prototype method. I have to rework the implementation to accommodate for that.

bajtos:
Fixed. @ritch PTAL.

If you consider the fix for `findById` as too hacky, then I can remove it and implement a `Model.beforeRemote` hook in loopback that will set the information in `ctx.req.remoteInstance` as appropriate.

bajtos:
@slnode test please

bajtos:
We have agreed with @ritch to expose `ctx.remotingContext` instead of these fine-grained properties.

ritch:
> We have agreed with @ritch to expose ctx.remotingContext instead of these fine-grained properties.

Edit: 

`req.remotingContext = ctx`

PR #202, opened 2015-04-14
Enable Travis CI for this project

bajtos:
Based on the discussion in #195, I would like to enable Travis CI for this project.

Benefits:

 - Anonymous access to build results - contributors see why their pull request failed the build
 - The build fails whenever `npm build` fails. This allows us to control the acceptance criteria, most notably the build will fail on linter errors.

Close #195

/to @ritch please review
/cc @rmg @raymondfeng @STRML

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,6 @@
+language: node_js
+node_js:
+  - "0.10"
+  - "0.12"
+  - "iojs"
+


bajtos:
Note that both [loopback](https://github.com/strongloop/loopback/blob/master/.travis.yml) and [loopback-datasource-juggler](https://github.com/strongloop/loopback-datasource-juggler/blob/master/.travis.yml) have Travis CI enabled.

ritch:
> Anonymous access to build results 

Isn't ci.strongloop.com open and linked from PRs?

ritch:
Anyways I like having a second source to verify that tests are actually failing instead of the CI which fails all the time for reasons unrelated to the PR.

ritch:
@bajtos can we add the badge along with this PR?

STRML:
Personally I can't access the CI without a user and pass, which I don't
have.

ritch:
> Personally I can't access the CI without a user and pass, which I don't
have.

Hmmm I did not know this. Thanks for mentioning it.

rmg:
@ritch our CI is private, it requires a GitHub account in the strongloop org.

bajtos:
> @ritch  can we add the badge along with this PR?

I am not sure if it's a good idea to add the badge. We used to have a problem in one of our other modules - it was not passing builds due to a timing/race condition in the tests, we didn't have time to fix, and thus the README was showing a red badge for quite some time. I know the right solution is to fix the build instead of hiding the badge, but sometimes the reality dictates different approach.

Also note that neither [loopback](https://github.com/strongloop/loopback) nor [loopback-datasource-juggler](https://github.com/strongloop/loopback-datasource-juggler) have the badge in their README.

@ritch Thoughts?

ritch:
@bajtos ok... I am fine merging this without the badge.

I'd like to see badges eventually though. Don't you?

PR #203, opened 2015-04-14
Fix typo introduced in fca2fb0

bajtos:
/to @ritch @STRML please review

### DIFF
--- a/test/rest.test.js
+++ b/test/rest.test.js
@@ -2206,7 +2206,7 @@ describe('strong-remoting-rest', function() {
 
       objects.afterError(method.name, function(ctx, next) {
         if (Array.isArray(hookContext)) {
-          hookContext.push(hookContext);
+          hookContext.push(ctx);
         } else if (typeof hookContext === 'object') {
           hookContext = [hookContext, ctx];
         } else {


test/rest.test.js
```diff
@@ -2206,7 +2206,7 @@ describe('strong-remoting-rest', function() {
```
ritch: This whole method could be a lot more clear. Why can't we init `hookContext` as an object and just set properties?

ritch: Also shouldn't we throw if the `afterError` handler is called multiple times?

bajtos: This implementation is based on what I have in juggler for operation hooks. It supports an arbitrary number of hook invocations, while keeping the test assertion safe from false positives.

Few examples:

expected 1 call got 0 calls - error message will be like "expected 'hook not called' to eql { foo: bar }"
(compare to your proposal - "expected {} to eql { foo: bar }")

expected 1 call got 2 calls - error message will be like "expected [{ foo: bar }, { foo: baz }] to eql { foo: bar }"
(I guess your proposal is better because we could throw a specific message)

expected 2 calls got 1 call - error message will be like "expected { foo: bar } to eql [{ foo: bar }, { foo: baz }]"
(your proposal does not support this case)

PR #204, opened 2015-04-17
Fix array coercion for non-array target type

bajtos:
Before this patch, when sending an array to an argument of type "object" or "any", the value was flattened and the first item was used instead of the array.

This patch fixes the issue by omitting recursive coersion when the target type is not an array.

See #196 that introduced the regression.

/to @ritch please review
/cc @STRML @XinV

### DIFF
--- a/lib/shared-method.js
+++ b/lib/shared-method.js
@@ -236,23 +236,24 @@ function escapeRegex(d) {
  * @return {*}                 Coerced argument.
  */
 function coerceAccepts(uarg, desc) {
-  // If array, invoke for each member of the array.
-  if (Array.isArray(uarg)) {
-    return uarg.map(function(arg) {
-      return coerceAccepts(arg, desc);
+  var name = desc.name || desc.arg;
+  var targetType = desc.type;
+  var targetIsArray = Array.isArray(targetType) && targetType.length === 1;
+
+  // If coercing an array to an erray,
+  // then coerce all members of the array too
+  if (targetIsArray && Array.isArray(uarg)) {
+    return uarg.map(function(arg, ix) {
+      // when coercing array items, use only name and type,
+      // ignore all other root settings like "required"
+      return coerceAccepts(arg, {
+        name: name + '[' + ix +']',
+        type: desc.type[0]
+      });
     });
   }
 
-  var name = desc.name || desc.arg;
-  var targetType = desc.type;
   var actualType = SharedMethod.getType(uarg);
-
-  // Target type may be an array. If so, expect the arg to be a member of that array,
-  // since we map this function over arrays.
-  var targetIsArray = Array.isArray(targetType) && targetType.length === 1;
-  if (targetIsArray) {
-    targetType = targetType[0];
-  }
 
   // is the arg optional?
   // arg was not provided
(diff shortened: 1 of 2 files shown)


STRML:
Aha. Thanks for the catch. :+1: from me.

test/rest.test.js
```diff
@@ -994,6 +994,37 @@ describe('strong-remoting-rest', function() {
         });
     });
 
+    it('should not flatten arrays for non-standard target type', function(done) {
+      var method = givenSharedStaticMethod(
+        function(arg, cb) { cb(null, { value: arg }); },
+        {
+          // Note: the string "array" is not a valid type
```
ritch: why is `array` not a valid type?

STRML: Wouldn't the right type be `['any']`?

lib/shared-method.js
```diff
@@ -236,24 +236,25 @@ function escapeRegex(d) {
...
+
+  // If coercing an array to an erray,
+  // then coerce all members of the array too
+  if (targetIsArray && Array.isArray(uarg)) {
+    return uarg.map(function(arg, ix) {
+      // when coercing array items, use only name and type,
+      // ignore all other root settings like "required"
+      return coerceAccepts(arg, {
+        name: name + '[' + ix +']',
+        type: desc.type[0]
```
ritch: this code is very hard to read ...I wonder if we need a `Description` object that has properties like `type` and `arrayValueType` or similar.

lib/shared-method.js
```diff
@@ -236,24 +236,25 @@ function escapeRegex(d) {
...
-    });
-  }
-
   var name = desc.name || desc.arg;
   var targetType = desc.type;
-  var actualType = SharedMethod.getType(uarg);
-
-  // Target type may be an array. If so, expect the arg to be a member of that array,
-  // since we map this function over arrays.
   var targetIsArray = Array.isArray(targetType) && targetType.length === 1;
```
ritch: this should really be named `targetIsTypedArray` or something that makes it clear that we are talking only about `['string']` and similar.

ritch:
Comments can be addressed elsewhere as they aren't really in the scope of this PR.

LGTM

lib/shared-method.js
STRML: `targetTypeIsArray` would be good imo - we don't want to say `TypedArray` which is a very different thing.

I agree, we can do cleanup in a separate PR.

test/rest.test.js
bajtos: See http://docs.strongloop.com/display/LB/LoopBack+types#LoopBacktypes-Arraytypes, these are the valid types:

 - `[String]`
 - `["String"]`
 - `[{type: String, length: 64}]`

In all cases, it's an array with a single item that describes the type of array items.

bajtos: For the purpose of the test, I can use `accepts: { arg: 'arg', type: 'any' }` if that's make the test easier to understand.

bajtos: See #205.

lib/shared-method.js
bajtos: See #205.

lib/shared-method.js
bajtos: I agree that such object would be great. However, I don't have time to make this refactoring now :(

PR #205, opened 2015-04-20
Cleanup in coercion code

bajtos:
- Use "any" instead of "array" in the unit test to make it easier
   to understand what's happening under the hood.

 - Rename "targetIsArray" to "targetTypeIsArray".

This is a follow-up for #204.

/to @ritch please review
/cc @STRML


ritch:
LGTM

PR #207, opened 2015-04-22
Tighten coercion and type checks in SharedMethod

bajtos:
- Modify `coerceAccepts()` to be more strict about the argument value matching the requested argument type.

- Modify the converter to return primitive types (e.g. "number 3") instead of object types (e.g. "Number(3)").

- Reject argument values that don't match target type even after the conversion was applied (HTTP status code 400 is returned in such case).

- Improve test failure messages by reporting errors from `object.invoke()` instead of failing on an assertion that expected successful invocation.

Connect to strongloop-internal/scrum-loopback#243

/to @ritch @raymondfeng please review
/cc @STRML

### DIFF
--- a/lib/shared-method.js
+++ b/lib/shared-method.js
@@ -237,7 +237,7 @@ function escapeRegex(d) {
  */
 function coerceAccepts(uarg, desc) {
   var name = desc.name || desc.arg;
-  var targetType = desc.type;
+  var targetType = convertToBasicRemotingType(desc.type);
   var targetIsArray = Array.isArray(targetType) && targetType.length === 1;
 
   // If coercing an array to an erray,
@@ -306,6 +306,39 @@ function coerceAccepts(uarg, desc) {
   }
 
   return uarg;
+}
+
+/**
+ * Returns an appropriate type based on a type specifier from remoting
+ * metadata.
+ * @param {Object} type A type specifier from remoting metadata,
+ *    e.g. "[Number]" or "MyModel" from `accepts[0].type`.
+ * @returns {String} A type name compatible with the values returned by
+ *   `SharedMethod.getType()`, e.g. "string" or "array".
+ */
+function convertToBasicRemotingType(type) {
+  if (Array.isArray(type)) {
+    return type.map(convertToBasicRemotingType);
+  }
+
+  if (typeof type === 'object') {
+    type = type.modelName || type.name;
+  }
+
+  type = String(type).toLowerCase();
+  switch (type) {
+    case 'string':
+    case 'number':
+    case 'date':
+    case 'boolean':
+    case 'buffer':
+    case 'object':
+    case 'any':
+      return type;
+    default:
+      // custom types like MyModel
+      return 'object';
+  }
 }
 
 function convertValueToTargetType(argName, value, targetType) {
--- a/lib/shared-method.js
+++ b/lib/shared-method.js
@@ -247,7 +247,7 @@ function coerceAccepts(uarg, desc) {
       // when coercing array items, use only name and type,
       // ignore all other root settings like "required"
       return coerceAccepts(arg, {
-        name: name + '[' + ix +']',
+        name: name + '[' + ix + ']',
         type: desc.type[0]
       });
     });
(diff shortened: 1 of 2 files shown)


lib/shared-method.js
```diff
@@ -247,7 +247,7 @@ function coerceAccepts(uarg, desc) {
       // when coercing array items, use only name and type,
       // ignore all other root settings like "required"
       return coerceAccepts(arg, {
-        name: name + '[' + ix +']',
+        name: name + '[' + ix + ']',
```
bajtos: This is fixing a linter issue.

lib/shared-method.js
```diff
@@ -269,35 +302,30 @@ function coerceAccepts(uarg, desc) {
...
-      return value;
+      return String(value).valueOf();
     case 'date':
       return new Date(value);
     case 'number':
-      return Number(value);
+      return Number(value).valueOf();
     case 'boolean':
-      return Boolean(value);
+      return Boolean(value).valueOf();
```
raymondfeng: What does the extra `valueOf` offer?

bajtos: It returns a primitive type instead of a boxed (wrapped) object.

- `typeof Boolean(true) === 'object'`
- `typeof (Boolean(true).valueOf()) === 'boolean`

STRML:
:+1: This is great.

lib/shared-method.js
```diff
@@ -255,8 +255,41 @@ function coerceAccepts(uarg, desc) {
...
+        name, targetType, uarg, typeof uarg, e.message);
+      throw new badArgumentError(message);
+    }
+  }
+
+  var typeMismatch = targetType !== 'any' &&
+    actualType !== 'undefined' &&
+    targetType !== actualType &&
+    // an array can be assigned to an object
+    !(targetType === 'object' && actualType === 'array');
```
STRML: Could you elaborate more on this (and put the explanation in the comment)?

bajtos: When you have an argument of `type: 'object'`, this line allows to assign an array value to this argument. In JavaScript, an array is an object too (`typeof [] === 'object'`). However, `SharedMethod.getType([])` returns `array` instead of `object`.

Does it make sense now? Should I include the whole paragraph above in the comment?

lib/shared-method.js
```diff
@@ -255,8 +255,41 @@ function coerceAccepts(uarg, desc) {
 
   var actualType = SharedMethod.getType(uarg);
 
-  // is the arg optional?
-  // arg was not provided
+  // convert values to the correct type
+  // TODO(bajtos) Move conversions to HttpContext (and friends)
+  // SharedMethod should only check that argument values match argument types.
```
bajtos: See https://github.com/strongloop/strong-remoting/issues/208

lib/shared-method.js
```diff
@@ -255,8 +255,41 @@ function coerceAccepts(uarg, desc) {
...
+
+  if (typeMismatch) {
+    var message = util.format('Invalid value for argument \'%s\' of type ' +
+      '\'%s\': %s. Received type was converted to %s.',
+      name, targetType, uarg, typeof uarg);
+    throw new badArgumentError(message);
+  }
+
+  // Verify that a required argument has a value
+  // FIXME(bajtos) "null" should be treated as no value too
```
bajtos: See https://github.com/strongloop/strong-remoting/issues/208

ritch:
LGTM

Lets wait until the CI for LoopBack passes using this version before release.

bajtos:
> Lets wait until the CI for LoopBack passes using this version before release.

I run the test myself to be sure, and I have discovered some issues indeed. See the additional commit and https://github.com/strongloop/loopback/pull/1332

@ritch PTAL again.

lib/shared-method.js
bajtos: @STRML ping

bajtos:
@ritch ping

> I run the test myself to be sure, and I have discovered some issues indeed. See the additional commit and strongloop/loopback#1332

lib/shared-method.js
```diff
@@ -269,35 +302,55 @@ function coerceAccepts(uarg, desc) {
...
+    case 'number':
+    case 'date':
+    case 'boolean':
+    case 'buffer':
+    case 'object':
+    case 'any':
+      return type;
+    default:
+      // custom types like MyModel
+      return 'object';
```
ritch: This might break complex type coercion. When you define a model in loopback we setup a strong-remoting type. Perhaps you can check that list here?

bajtos: > This might break complex type coercion. 

Can you please describe such scenario, so that I can add a unit-test for that?

> Perhaps you can check that list here?

To be honest, I feel it's out of scope of this pull request. If the data read from the request is an object, and the argument type is object-like (either object, any, or custom type from LoopBack), then I think it's ok to proceed.

If we start checking constructor names, then we might break applications that are working now.

bajtos: Can we defer the change you are proposing and implement it as part of #208?

bajtos: Also note that this function must return the same set of type strings as [SharedMethod.getType](https://github.com/strongloop/strong-remoting/blob/97a65d095addedf73def2a01204890b865880216/lib/shared-method.js#L363-L397), because that's the other side in the comparison.

```js
if (convertToBasicRemotingType(desc.type) !== SharedMethod.getType(argValue)) {
  // type mismatch
}
```

ritch: Here is the scenario I have in mind... Actually I don't think this is an issue. Can you confirm?

```js
Person.prototype.greet = function(msg, someone, cb) {
  this.log('Greetings! - ' + msg);
  someone.greet('hello ' + this.name);
  cb();
}

Person.remoteMethod('greet', {
  isStatic: false,
  accepts: [{type: 'string'}, {type: 'Person'}]
});

Person.prototype.log = function(msg) {
  console.log(this.name + ' says: ' + msg);
}
```

bajtos: Yes, that should work. In fact, I believe LoopBack built-in models are using this and their tests are passing with this patch in place.

ritch:
Ok I think we can merge this now. :+1: LGTM

bajtos:
For posterity, I have rebased the commit on top of the latest master and added two more comment sections to capture our discussions above.

lib/shared-method.js
elankeeran: typeof Boolean(true) returns "boolean" primitive type then what does 'wrapped object'?

STRML: Yeah, that's incorrect on all three lines - `new Boolean(true)` returns a wrapped object, but `Boolean(true)` does not. Same with String & Number.

bajtos: Good catch. Would you mind submitting a pull request to fix this? Is there a way how to write a test that fails with the current implementation and passes with the fix applied?

STRML: No, there isn't. The lines are equivalent. With the `valueOf()` on my machine, I get 27M iterations; without, 44M. It won't matter either way.

bajtos: Makes sense. Let's create a pull request with no tests then. Any volunteers?

PR #209, opened 2015-04-28
add test to verify target type array

PradnyaBaviskar:
@bajtos, 
In reference to https://github.com/strongloop/loopback/issues/1327, ```accepts: { arg: 'arg', type: 'array' }```, worked out of the box, without me requiring to make any changes!

I've just added test cases to verify target type ```array```

Can you please review?

### DIFF
--- a/lib/shared-method.js
+++ b/lib/shared-method.js
@@ -257,7 +257,7 @@ function coerceAccepts(uarg, desc) {
       // ignore all other root settings like "required"
       return coerceAccepts(arg, {
         name: name + '[' + ix + ']',
-        type: desc.type[0]
+        type: targetType[0]
       });
     });
   }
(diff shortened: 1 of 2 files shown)


bajtos:
Yes, it works out of the box because unknown type is treated as "any" or "object" at the moment. Unfortunately all users of remoting metadata must be aware of this behaviour, that's why I was proposing to convert `type: 'array'` to `type: ['any']`.

PradnyaBaviskar:
I've made corresponding changes. Can you review now?

bajtos:
@PradnyaBaviskar I figured out it will be easier to make the change as I have envisioned it, rather than trying to explain it. See #210.

I think the tests you have added are still useful, can you please rebase the code after #210 is landed?

It would be nice to add a test that an argument of type `array` accepts an array containing items of different types, e.g. a number and a string: `['text', 12345]`

bajtos:
@PradnyaBaviskar is pull request this still relevant? Could you please rebase it on top of the current master?

PradnyaBaviskar:
Rebased.

bajtos:
Landed, thank you :bow:

PR #210, opened 2015-05-04
Support arg type "array"

bajtos:
Support `type: 'array'` as a syntactic sugar for `type: ['any']`.

Connect to strongloop/loopback#1327
See also #209 

/to @ritch please review
/cc @PradnyaBaviskar @fabien


ritch:
LGTM

PR #212, opened 2015-05-06
Add authorization hook

ritch:
/to @raymondfeng 
/cc @fabien @bajtos 

This pull request adds a new hook that is executed in the following order:

1. invokeMethodInContext
2. **authorization hook**
3. before hooks
4. method
5. after hooks

Here is a basic example implementing the hook:

```js
remotes.authorization = function(ctx, next) {
  // you must callback with an error to cancel the invokation
  next(new Error('denied'));
}
```

### DIFF
--- a/lib/remote-objects.js
+++ b/lib/remote-objects.js
@@ -577,6 +577,16 @@ function searchListenerTree(handlers, type, tree, i) {
   return listeners;
 }
 
+RemoteObjects.prototype._executeAuthorizationHook = function(ctx, cb) {
+  if (typeof this.authorization === 'function') {
+    this.authorization(ctx, cb);
+  } else {
+    process.nextTick(function() {
+      cb();
+    });
+  }
+};
+
 /**
  * Invoke the given shared method using the supplied context.
  * Execute registered before/after hooks.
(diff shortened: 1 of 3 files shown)


lib/remote-objects.js
```diff
@@ -577,6 +577,16 @@ function searchListenerTree(handlers, type, tree, i) {
   return listeners;
 }
 
+RemoteObjects.prototype._executeAuthorizationHook = function(ctx, cb) {
+  if(typeof this.authorization === 'function') {
+    this.authorization(ctx, cb);
+  } else {
+    process.nextTick(function() {
```
bajtos: Is this necessary? AFAIK, `process.nextTick` slows down server throughput, at least in benchmarks.

test/authorize-hook.test.js
```diff
@@ -0,0 +1,52 @@
...
+      invokeRemote(server.address().port,
+        function(err, session) {
+          expect(err).to.not.exist();
+          expect(session.userId).to.equal(123);
+          expect(called).to.equal(true);
+          done();
+        }
+      );
+    });
+  });
```
bajtos: Please add a unit-test to verify that the authorization hook is called before any other hooks. That's the goal of this patch, right?

bajtos:
The CI build failed, please fix it.

test/authorize-hook.test.js
ritch: Used your callStack array trick. Works like a charm. Though the confusing part is that since we are using `remotes.connect()` and `remotes.invoke()` the local before hook is also included in the `callStack`.

raymondfeng:
LGTM

Issue #214, opened 2015-06-10
Support returning a transport agnostic ReadableStream


ritch:
Here is a basic example of what I'd like to be able to do:

```js
function createStream(options, cb) {
  var stream = createReadableObjectStream();
  cb(null, stream);
}

sharedClass.defineMethod('createStream', {
  returns: {
    arg: 'stream',
    type: 'ReadableStream',
    json: true,
    contentType: 'application/json; boundary=NL' // this is the default
  }
});
```

For adapters that support multiple encodings (http) we would need a way to specify the required info.

/cc @raymondfeng @bajtos

bajtos:
FWIW, I think this is a duplicate of #35.

> For adapters that support multiple encodings (http) we would need a way to specify the required info.

The discussion in #35 has some ideas how this could be implemented.

gausie:
Will this support setting result headers programmatically? 

I currently need to set the name of the file that is being streamed based on model data. I do so by hacking `res` in as a parameter (using `accepts: [ { arg: 'res', type: 'object', http: { source: 'res' } } ]`) then using `res.setHeader` and finally manually piping my stream to res.

Ideally in closing this bug, it would be made possible to achieve this as well!

ritch:
@gausie you can do this now by setting a return argument to `{http: {target: 'header', header: 'Content-Type': myDynamicContentType}`. But I think we can make this a bit better by using the streams encoding property or another property that you can just set on the stream itself (so the function stays transport agnostic).

I'd suggest creating a new issue which explains your use case so others can chime in and we can up the priority.

richardpringle:
Apparently this issue has a couple of duplicates #35, strongloop/loopback-swagger#34, and strongloop/loopback#2209... and I'm realizing now that it leads to a bit of a wild goose chase. I am going to close this issue in favor of the others and I believe that his issue actually may have been solved already by #216. 

Please re-open and mention me directly if there are still concerns!

PR #216, opened 2015-06-11
Add object mode ReadableStream support

ritch:
This PR adds support for returning object mode `ReadableStream`s.

connect to strongloop/strong-remoting#214

Using HTTP `ReadableStream`s are piped to the `res` using https://github.com/dominictarr/mux-demux. This allows for `stream.on('error')`s to be sent using the multiplexed NL separated JSON protocol.

I would still like to add the same exact capability using multipart responses.

/to @bajtos 
/cc @raymondfeng @kraman @rmg (noticed you had used mux-demux)

### DIFF
--- a/lib/http-context.js
+++ b/lib/http-context.js
@@ -23,6 +23,7 @@ var DEFAULT_SUPPORTED_TYPES = [
   ];
 
 var MuxDemux = require('mux-demux');
+var SSEClient = require('sse').Client;
 
 /**
  * Create a new `HttpContext` with the given `options`.
@@ -41,6 +42,10 @@ function HttpContext(req, res, method, options) {
   this.methodString = method.stringName;
   this.supportedTypes = this.options.supportedTypes || DEFAULT_SUPPORTED_TYPES;
   this.result = {};
+
+  var streamsDesc = method.streams;
+  var returnStreamDesc = streamsDesc && streamsDesc.returns;
+  var methodReturnsStream = !!returnStreamDesc;
 
   if (this.supportedTypes === DEFAULT_SUPPORTED_TYPES && !this.options.xml) {
     // Disable all XML-based types by default
@@ -64,18 +69,26 @@ function HttpContext(req, res, method, options) {
 inherits(HttpContext, EventEmitter);
 
 HttpContext.prototype.createStream = function() {
-  var streamDesc = this.method.stream;
+  var streamsDesc = this.method.streams;
+  var returnStreamDesc = streamsDesc && streamsDesc.returns;
   var mdm = this.muxDemuxStream = new MuxDemux();
   var io = this.io = {};
   var res = this.res;
 
   debug('create stream');
 
-  if (streamDesc.json && streamDesc.type === 'ReadableStream') {
-    this.io.out = mdm.createWriteStream();
-    // since the method returns a ReadableStream
-    // setup an output to pipe the ReadableStream to
-    mdm.pipe(res);
+  if (returnStreamDesc.json && returnStreamDesc.type === 'ReadableStream') {
+
+    if (!this.shouldReturnEventStream()) {
+      res.setHeader('Content-Type',
+      returnStreamDesc.contentType || 'application/json; boundary=NL');
+      res.setHeader('Transfer-Encoding', 'chunked');
+
+      this.io.out = mdm.createWriteStream();
+      // since the method returns a ReadableStream
+      // setup an output to pipe the ReadableStream to
+      mdm.pipe(res);
+    }
   }
 };
 
@@ -436,29 +449,66 @@ function toXML(input) {
   }
   return xml;
 }
+
+HttpContext.prototype.shouldReturnEventStream = function() {
+  var req = this.req;
+  var query = req.query;
+  var format = query._format;
+
+  var acceptable = req.accepts('text/event-stream');
+  return (format === 'event-stream') || acceptable;
+};
+
+HttpContext.prototype.respondWithEventStream = function(stream) {
+  var client = new SSEClient(this.req, this.res);
+
+  client.initialize();
+
+  stream.on('data', function(chunk) {
+    client.send('data', JSON.stringify(chunk));
+  });
+
+  stream.on('error', function(err) {
+    var outErr = {message: err.message};
+    for (var key in err) {
+      outErr[key] = err[key];
+    }
+    client.send('error', JSON.stringify(outErr));
+  });
+
+  stream.on('end', function() {
+    client.send({event: 'end', data: 'null'});
+  });
+};
+
 /**
  * Finish the request and send the correct response.
  */
 
 HttpContext.prototype.done = function(cb) {
+  var ctx = this;
   var method = this.method;
-  var streamDesc = method.stream;
-  var methodReturnsStream = !!streamDesc;
+  var streamsDesc = method.streams;
+  var returnStreamDesc = streamsDesc && streamsDesc.returns;
+  var methodReturnsStream = !!returnStreamDesc;
   var res = this.res;
   var out = this.io && this.io.out;
   var err = this.error;
   var result = this.result;
 
   if (methodReturnsStream) {
-    if (streamDesc.json) {
+    if (returnStreamDesc.json) {
       debug('handling json stream');
-      res.setHeader('Content-Type',
-      streamDesc.contentType || 'application/json; boundary=NL');
-      res.setHeader('Transfer-Encoding', 'chunked');
 
-      var stream = result[streamDesc.arg];
+      var stream = result[returnStreamDesc.arg];
 
-      if (streamDesc.type === 'ReadableStream') {
+      if (returnStreamDesc.type === 'ReadableStream') {
+        if (ctx.shouldReturnEventStream()) {
+          debug('respondWithEventStream');
+          ctx.respondWithEventStream(stream);
+          return;
+        }
+
         debug('piping to mdm stream');
         stream.pipe(out);
         stream.on('error', function(err) {
@@ -474,7 +524,7 @@ HttpContext.prototype.done = function(cb) {
         });
         // TODO(ritch) support multi-part streams
       } else {
-        cb(new Error('unsupported stream type: ' + streamDesc.type));
+        cb(new Error('unsupported stream type: ' + returnStreamDesc.type));
       }
     } else {
       cb(new Error('Unsupported stream descriptor, only descriptors ' +
--- a/lib/http-context.js
+++ b/lib/http-context.js
@@ -22,6 +22,8 @@ var DEFAULT_SUPPORTED_TYPES = [
     '*/*'
   ];
 
+var MuxDemux = require('mux-demux');
+
 /**
  * Create a new `HttpContext` with the given `options`.
  *
@@ -48,6 +50,11 @@ function HttpContext(req, res, method, options) {
   }
 
   req.remotingContext = this;
+
+  // streaming support
+  if (methodReturnsStream) {
+    this.createStream();
+  }
 }
 
 /**
@@ -55,6 +62,22 @@ function HttpContext(req, res, method, options) {
  */
 
 inherits(HttpContext, EventEmitter);
+
+HttpContext.prototype.createStream = function() {
+  var streamDesc = this.method.stream;
+  var mdm = this.muxDemuxStream = new MuxDemux();
+  var io = this.io = {};
+  var res = this.res;
+
+  debug('create stream');
+
+  if (streamDesc.json && streamDesc.type === 'ReadableStream') {
+    this.io.out = mdm.createWriteStream();
+    // since the method returns a ReadableStream
+    // setup an output to pipe the ReadableStream to
+    mdm.pipe(res);
+  }
+};
 
 /**
  * Build args object from the http context's `req` and `res`.
@@ -417,11 +440,52 @@ function toXML(input) {
  * Finish the request and send the correct response.
  */
 
-HttpContext.prototype.done = function() {
+HttpContext.prototype.done = function(cb) {
+  var method = this.method;
+  var streamDesc = method.stream;
+  var methodReturnsStream = !!streamDesc;
+  var res = this.res;
+  var out = this.io && this.io.out;
+  var err = this.error;
+  var result = this.result;
+
+  if (methodReturnsStream) {
+    if (streamDesc.json) {
+      debug('handling json stream');
+      res.setHeader('Content-Type',
+      streamDesc.contentType || 'application/json; boundary=NL');
+      res.setHeader('Transfer-Encoding', 'chunked');
+
+      var stream = result[streamDesc.arg];
+
+      if (streamDesc.type === 'ReadableStream') {
+        debug('piping to mdm stream');
+        stream.pipe(out);
+        stream.on('error', function(err) {
+          var outErr = {message: err.message};
+          for (var key in err) {
+            outErr[key] = err[key];
+          }
+
+          // this is the reason we are using mux-demux
+          out.error(outErr);
+
+          out.end();
+        });
+        // TODO(ritch) support multi-part streams
+      } else {
+        cb(new Error('unsupported stream type: ' + streamDesc.type));
+      }
+    } else {
+      cb(new Error('Unsupported stream descriptor, only descriptors ' +
+        'with property "json:true" are supported'));
+    }
+    return;
+  }
+
   // send the result back as
   // the requested content type
   var data = this.result;
-  var res = this.res;
   var accepts = this.req.accepts(this.supportedTypes);
   var defaultStatus = this.method.http.status;
 
@@ -479,4 +543,6 @@ HttpContext.prototype.done = function() {
     }
     res.end();
   }
+
+  cb();
 };
--- a/lib/http-invocation.js
+++ b/lib/http-invocation.js
@@ -19,6 +19,8 @@ var Dynamic = require('./dynamic');
 var SUPPORTED_TYPES = ['json', 'application/javascript', 'text/javascript'];
 var qs = require('qs');
 var urlUtil = require('url');
+var ReadableStream = require('stream').Readable;
+var MuxDemux = require('mux-demux');
 
 /*!
  * JSON Types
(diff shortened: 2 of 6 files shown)


lib/http-context.js
```diff
@@ -304,6 +306,31 @@ HttpContext.prototype.invoke = function(scope, method, fn, isCtor) {
...
+        res.setHeader('Content-Type',
+          streamDesc.contentType || 'application/json; boundary=NL');
+        res.setHeader('Transfer-Encoding', 'chunked');
+
+        var stream = results[streamDesc.arg];
+
+        if (streamDesc.type === 'ReadableStream') {
+          stream.pipe(JSONStream.stringify(false)).pipe(res);
+        } else {
+          fn(new Error('unsupported stream type'));
```
bajtos: Please include the actual type value in the error message. If you think it's a sensitive info that should not be returned to the client, then print a debug log. The goal is to simplify troubleshooting.

lib/http-context.js
```diff
@@ -304,6 +306,31 @@ HttpContext.prototype.invoke = function(scope, method, fn, isCtor) {
...
+        }
+      } else {
+        fn(new Error('unsupported stream description'));
```
bajtos: Please provide more information to help the user to understand better what they did wrong and how to fix the problem.

Example: `Unsupported stream descriptor, only descriptors with property "json:true" are supported.`.

lib/http-context.js
```diff
@@ -379,6 +406,11 @@ function toXML(input) {
  */
 
 HttpContext.prototype.done = function() {
+  if (this.method.stream) {
+    // the stream has already been setup
+    return;
+  }
+
```
bajtos: Is there any particular reason why you cannot handle streams in this `done` method? To me, it seems more logical to handle all different response arguments (like body stream) in a single place.

lib/http-context.js
```diff
@@ -304,6 +306,31 @@ HttpContext.prototype.invoke = function(scope, method, fn, isCtor) {
   var pipe = http && http.pipe;
   var pipeDest = pipe && pipe.dest;
```
bajtos: I am proposing to deprecate `http.pipe.dest` in favour of this new "returns stream" API. Thoughts?

lib/http-invocation.js
```diff
@@ -223,14 +228,30 @@ HttpInvocation.prototype.invoke = function(callback) {
...
+  var streamDesc = method.sharedMethod.stream;
+  var methodReturnsStream = !!streamDesc;
+  var streamResult;
+
+  if (methodReturnsStream) {
+    if (streamDesc.type === 'ReadableStream' && streamDesc.json) {
+      streamResult = request(this.req).pipe(JSONStream.parse());
+      callback(null, streamResult);
+    } else {
+      callback(new Error('unsupported stream type'));
```
bajtos: Please add more info to the error message, make it easier to troubleshoot the problem.

test/streams.js
```diff
@@ -45,3 +47,66 @@ describe('strong-remoting', function() {
...
+      var buf = [];
+      var n = 5;
+      while (n--) {
+        buf.push({n: n});
+      }
+      rs._read = function() {
+        setTimeout(function() {
+          rs.push(buf.shift() || null);
+        }, 1);
+      };
```
bajtos: How about using even-stream's [readArray](https://www.npmjs.com/package/event-stream#readarray-array)? Unless the 1ms timeout is important for your test.

test/streams.js
```diff
@@ -45,3 +47,66 @@ describe('strong-remoting', function() {
...
+    remotes.invoke('StreamClass.createStream', [], function(err, stream) {
+      assert(stream.readable, 'must be a ReadableStream');
+      var count = 0;
+      stream.on('data', function(d) {
+        count++;
+
+        if (count === 5) {
+          done();
+        }
+      });
```
bajtos: Perhaps use [writeArray](https://www.npmjs.com/package/event-stream#writearray-callback) and assert on the array content?

lib/http-context.js
ritch: Ya... I'll add an issue

lib/http-context.js
ritch: We need to handle errors and I can't reference the error here (unless there is a `ctx.error` property??).

test/streams.js
ritch: Good idea. The `setTimeout` is to add some delay, but I don't think its strictly necessary.

test/streams.js
ritch: Good idea... will do!

rmg:
Missing `JSONStream` dependency.. also, please consider switching to `jsonstream` instead - it's the canonical version of the module. Sadly, since OS X uses a case-folding filesystem, it can't actually tell the difference between them :-(

lib/http-context.js
```diff
@@ -38,6 +41,8 @@ function HttpContext(req, res, method, options) {
   this.args = this.buildArgs(method);
   this.methodString = method.stringName;
   this.supportedTypes = this.options.supportedTypes || DEFAULT_SUPPORTED_TYPES;
+  var streamDesc = method.stream;
+  var methodReturnsStream = !!streamDesc;
```
bajtos: IIUC, the stream mode is turned on by marking a return argument as `type: 'stream'`. This code seems to check the method metadata instead?

test/streams.js
```diff
@@ -45,3 +48,87 @@ describe('strong-remoting', function() {
...
+describe('a function returning a ReadableStream', function() {
+  var Readable = require('stream').Readable;
+  var remotes = RemoteObjects.create();
+  var streamClass;
+  var server;
+
+  before(function(done) {
+    var test = this;
+    // NOTE: Date object will not be coerced back into a Date
+    var data = test.data = [{foo: 'bar'}, 'bat', false, 0, null];
```
bajtos: I find the comment confusing, because I don't see any Date object in the data?

test/streams.js
```diff
@@ -45,3 +48,87 @@ describe('strong-remoting', function() {
...
+
+    streamClass = new SharedClass('StreamClass', StreamClass);
+
+    streamClass.defineMethod('createStream', {
+      isStatic: true,
+      fn: StreamClass.createStream,
+      returns: [{
+        arg: 'stream',
+        type: 'ReadableStream',
+        json: true
```
bajtos: Should we specify `root:true` as well? Or is that implied by `type:ReadableStream`? Could you please use a different arg name in the test, just to make sure that we don't accidentally match it when deciding whether the argument is of a stream type?

bajtos:
@ritch do you expect any non-javascript consumers of this new stream API? If you do, then could you please add a unit-test showing how to read and parse the stream using plain REST?

The Angular code generator will not support this new stream type OOTB, do you have any plans for that SDK? Should it ignore (skip) any methods that return a stream?

test/streams.js
bajtos: I mean use for example `arg: 'result'`' instead of `arg: 'stream'`.

lib/shared-method.js
```diff
@@ -82,6 +82,7 @@ function SharedMethod(fn, name, sc, options) {
   this.documented = (options.documented || fn.documented) !== false;
   this.http = options.http || fn.http || {};
   this.rest = options.rest || fn.rest || {};
+  this.authorization = options.authorization;
```
bajtos: This seems like an unrelated change?

lib/shared-method.js
```diff
@@ -109,6 +110,13 @@ function SharedMethod(fn, name, sc, options) {
   }
   this.returns.forEach(normalizeArgumentDescriptor);
 
+  var firstReturns = this.returns[0];
+  var streamTypes = ['ReadableStream', 'WriteableStream', 'DuplexStream'];
+
+  if (firstReturns && firstReturns.type && streamTypes.indexOf(firstReturns.type) > -1) {
+    this.stream = firstReturns;
```
bajtos: I'd like to eventually add support for streams to input arguments too. Would you mind using a different property name to distinguish between "accepts" and "returns" streams? E.g. "streamResult" or "streamReturn"

lib/http-context.js
bajtos: Never mind, I see that `method.stream` is set in shared-method.js. I'll delete these two comments.

test/streams.js
ritch: The test would fail with a `Date`

lib/shared-method.js
ritch: Hmmm... that is odd...?

lib/http-invocation.js
```diff
@@ -223,14 +229,37 @@ HttpInvocation.prototype.invoke = function(callback) {
   if (!this.req) {
     this.createRequest();
   }
-  request(this.req, function(err, res, body) {
-    if (err instanceof SyntaxError) {
-      if (res.status === 204) err = null;
+  var method = this.method;
+  var streamDesc = method.sharedMethod.stream;
+  var methodReturnsStream = !!streamDesc;
+  // var streamResult;
```
bajtos: please remove commented-out code.

lib/http-invocation.js
```diff
@@ -223,14 +229,37 @@ HttpInvocation.prototype.invoke = function(callback) {
...
+  if (methodReturnsStream) {
+    if (streamDesc.type === 'ReadableStream' && streamDesc.json) {
+      var mdm = new MuxDemux();
+      mdm.on('connection', function(stream) {
+        callback(null, stream);
+      });
+
+      request(this.req).pipe(mdm);
+
+      // streamResult = request(this.req).pipe(JSONStream.parse());
```
bajtos: please remove commented-out code.

lib/http-invocation.js
```diff
@@ -223,14 +229,37 @@ HttpInvocation.prototype.invoke = function(callback) {
...
+    } else {
+      callback(new Error('unsupported stream type'));
     }
-    if (err) return callback(err);
-    self.res = res;
-    self.transformResponse(res, body, callback);
-  });
+
+    return;
+  } else {
```
bajtos: It may be my personal preference, but I would omit unnecessary `else` block after `return` and thus keep the amount of whitespace changes lower.

test/streams.js
bajtos: I see, could you please expand the comment to explain that the test data includes different object types, but that Date is intentionally excluded?

It would be good to document this limitation somewhere, or even emit a warning.

bajtos:
@ritch I am assigning this back to you, please let me know when the code is ready for the final review.

test/streams.js
ritch: > Should we specify root:true as well? Or is that implied by type:ReadableStream

That is implied by `type: ReadableStream`

ritch: Changed the arg to `result`

lib/http-invocation.js
ritch: removed

lib/http-invocation.js
ritch: omitted

lib/shared-method.js
ritch: removed

lib/shared-method.js
ritch: renamed to `streams` it includes (right now) a `returns` property for the first argument in the callback that is a stream

lib/http-invocation.js
ritch: removed

ritch:
This should be ready to merge @bajtos

bajtos:
@ritch the CI is red (both Travis and Jenkins), PTAL.

lib/http-context.js
```diff
@@ -374,15 +405,97 @@ function toXML(input) {
...
+          out.end();
+        });
+        // TODO(ritch) support multi-part streams
+      } else {
+        cb(new Error('unsupported stream type: ' + returnStreamDesc.type));
+      }
+    } else {
+      cb(new Error('Unsupported stream descriptor, only descriptors ' +
+        'with property "json:true" are supported'));
+    }
```
bajtos: Nitpick: I would personally move the contents of the block `if(methodReturnsStream)` to a new method, e.g. `HttpContext.prototype._streamResponse`. The purpose is to keep the `done` method reasonably short.

ritch:
test please

bajtos:
@ritch It seems to me that the commit "Merge and fix iojs failures" is bringing in a bunch of changes already landed on the master branch. Would you mind rebasing this PR on the current master, so that 1) the history is linear 2) the diff on github is not polluted with changes that were already reviewed and landed?

ritch:
@bajtos I'm not going to rewrite the history. It takes forever and is error prone (since the commits conflict on the package deps and dev deps.

This PR has gone on way longer than is practical. Just look at the diff instead of the commits.

ritch:
@bajtos ...actually let me show you the pain I am going through when rebasing and perhaps you can either see why I am refusing... or you might be able to see what I am doing incorrectly and prevent future pain.

Here is what I am doing:

```
git fetch # ensure we have latest from master
git rebase master -i # so we can squash the commits

# outputs...
pick 6d9ba85 Add object mode ReadableStream support
pick 4038439 Add support for Event Source streams

# I change the second "pick" to "s" to squash it into the previous commit
# note the merge commit doesn't show up...

# but git is unable to perform the rebase:

could not apply 6d9ba85... Add object mode ReadableStream support
```

If I don't try to squash the commit, the rebase outputs:

```
First, rewinding head to replay your work on top of it...
Applying: Add object mode ReadableStream support
Using index info to reconstruct a base tree...
M	lib/http-context.js
M	lib/rest-adapter.js
M	lib/shared-method.js
M	package.json
M	test/streams.js
Falling back to patching base and 3-way merge...
Auto-merging test/streams.js
Auto-merging package.json
**CONFLICT (content): Merge conflict in package.json**
Auto-merging lib/shared-method.js
Auto-merging lib/rest-adapter.js
Auto-merging lib/http-context.js
**CONFLICT (content): Merge conflict in lib/http-context.js**
Failed to merge in the changes.
Patch failed at 0001 Add object mode ReadableStream support
```

Which I can resolve manually. Its just pretty tedious and easy to merge incorrectly since the diff looks like this:

```
  "dependencies": {
<<<<<<< HEAD
    "express": "4.x",
    "body-parser": "^1.12.4",
    "debug": "^2.2.0",
    "eventemitter2": "^0.4.14",
    "cors": "^2.6.0",
    "jayson": "^1.2.0",
    "js2xmlparser": "^0.1.9",
    "async": "^0.9.0",
    "traverse": "^0.6.6",
    "request": "^2.55.0",
    "qs": "^2.4.2",
    "inflection": "^1.7.1",
    "xml2js": "^0.4.8"
  },
  "devDependencies": {
    "bluebird": "^2.9.25",
    "browserify": "~10.2.0",
    "chai": "^2.3.0",
=======
    "async": "^0.9.0",
    "body-parser": "^1.7.0",
    "cors": "^2.5.2",
    "debug": "^2.0.0",
    "eventemitter2": "^0.4.14",
    "express": "4.x",
    "inflection": "^1.4.2",
    "jayson": "^1.1.1",
    "js2xmlparser": "^0.1.3",
    "mux-demux": "^3.7.9",
    "qs": "^2.2.3",
    "request": "^2.42.0",
    "traverse": "^0.6.6",
    "xml2js": "^0.4.4"
  },
  "devDependencies": {
    "bluebird": "^2.9.6",
    "browserify": "~5.11.1",
    "chai": "^1.10.0",
    "event-stream": "^3.3.1",
>>>>>>> Add object mode ReadableStream support
```

Anyways... I don't need help rebasing... I can get all of that to work. Also I am not against rebasing in general or rewriting history. I'm just tired of re-merging the same changes over and over again only to have missed a version number, have that fail the build on ci and then spend time finding that mistake.

So if you have any way to make this a bit easier or less error prone I would love to know.

bajtos:
> So if you have any way to make this a bit easier or less error prone I would love to know.

I am afraid I don't have any better solution, I think this is the inherent cost of long-living feature branches :(

> Just look at the diff instead of the commits. 

That's what I did. The diff on github includes many unrelated changes, so it's rather difficult to distinguish signal from noise. Fortunately git client shows shorter diff, I'll review using that way.

package.json
```diff
@@ -13,24 +13,28 @@
...
-    "qs": "^2.4.2",
-    "inflection": "^1.7.1",
-    "xml2js": "^0.4.8"
+    "xml2js": "^0.4.4"
   },
   "devDependencies": {
-    "bluebird": "^2.9.25",
-    "browserify": "~10.2.0",
-    "chai": "^2.3.0",
+    "bluebird": "^2.9.6",
```
bajtos: AFAICT, you are downgrading the min bluebird version from 2.9.25 to 2.9.6, is that intentional?

package.json
```diff
@@ -13,24 +13,28 @@
...
+    "browserify": "~5.11.1",
```
bajtos: downgrade from ~10.2.0 to ~5.11.1, is it intentional?

bajtos:
Your package.json changes are downgrading certain dependencies, I pointed out two cases in my comments above. Could you please double-check other deps too?

bajtos:
The code looks good otherwise, merge at will.

rmg:
The rebase conflicts in `lib/http-context.js` are because of the merging of master into the branch instead of frequent rebasing to keep up.

The rebase conflicts in `package.json` are because master doesn't list the dependencies in alphabetical order but your commit that added 2 dependencies re-ordered them. This is where `git add -p` is your friend, since it allows you to commit the re-ordering as a separate commit from adding the deps.

As for re-writing history... A PR is a patch set proposal, and rewriting its history is more akin to working with an editor and revising a manuscript before sending it to the printers where it _does_ become part of the canonical history. This is where rebasing is a good thing.

package.json
rmg: Downgraded because the merge from master discarded the change from master.

package.json
rmg: Downgraded because the merge from master discarded the change from master.

rmg:
The numerous dependency downgrades are the result of the merge commit discarding the changes to package.json that were made to master since this branch started.

bajtos:
It looks to me like not rewriting the history is even more error prone that rewriting it.

ritch:
So I guess rewriting history itself is not the issue. Its manual merging. Merging is error prone and tedious. It sounds like the only way to avoid having to manually merge files is to rebase often (thanks for pointing that out @rmg).

ritch:
test please

Issue #186, opened 2015-03-17
Declarative pattern for response codes


dashby3000:
In order to return a custom non-error response code, you must implement an afterRemoting hook.  The following is an example of an endpoint returning a status of 299.

```javascript
MyModel.remoteMethod(
   'foo',
   {
     description: 'Do some foo',
     accepts: [
     ],
     returns:{ root: true} ,
     http: {
       verb: 'GET',
       path: '/foo'
     },
     rest: {
       after: function(ctx, cb) {
         ctx.res.status(299);
         cb();
       }
     }
   }
 );
```

We request a more declarative approach.  Something like: `returns: { status: 299, root: true }`

ref: PLATAPI-234

This would be the mechanism for implementing https://github.com/strongloop/loopback/issues/360

ritch:
Yeah this sounds good... It belongs in the `http` namespace though.

bajtos:
Another option to consider: annotate a callback argument as holding the HTTP status code 

```js
function foo(cb) {
  cb(null, 299);
}
foo.returns = {
  arg: 'httpStatusCode', http: { target: 'statusCode' }
}
```

ritch:
https://github.com/strongloop/strong-remoting/pull/218

ritch:
Released as **v2.19.0**

PR #218, opened 2015-06-16
Add support for declaring status code defaults

ritch:
/to @bajtos 
/cc @raymondfeng @dashby3000 

This adds a few new features:

 - declare a default `status` and `errorStatus` code in your `method.http` options
 - `returns.http.target`
   - `status` sets the `res.statusCode` to the provided value
   - `header` sets the `returns.header` or `returns.arg` named header to the value
 - set a custom header name for a header arg: `returns: {arg: 'length', {http: {target: 'header', header: 'Content-Length'}}}`
 
```js
describe('using a custom satus code', function() {
  it('returns a custom status code', function(done) {
    var method = givenSharedStaticMethod(
      function fn(cb) {
        cb();
      },
      {
        http: { status: 201 }
      }
    );
    json(method.url)
      .expect(201, done);
  });
  it('returns a custom error status code', function(done) {
    var method = givenSharedStaticMethod(
      function fn(cb) {
        cb(new Error('test error'));
      },
      {
        http: { status: 201, errorStatus: 508 }
      }
    );
    json(method.url)
      .expect(508, done);
  });
  it('returns a custom status code from a callback arg', function(done) {
    var exampleStatus = 222;
    var method = givenSharedStaticMethod(
      function fn(status, cb) {
        cb(null, status);
      },
      {
        accepts: { arg: 'status', type: 'number' },
        returns: {
          arg: 'status',
          http: { target: 'status' }
        }
      }
    );
    json(method.url + '?status=' + exampleStatus)
      .expect(exampleStatus, done);
  });
});
```

Connect to strongloop/strong-remoting#186

### DIFF
--- a/lib/http-context.js
+++ b/lib/http-context.js
@@ -340,7 +340,7 @@ HttpContext.prototype.invoke = function(scope, method, fn, isCtor) {
       if (err) {
         if (defaultErrorStatus &&
           (res.statusCode === undefined || res.statusCode === 200)) {
-          res.status(defaultErrorStatus);
+          res.status(err.status || err.statusCode || defaultErrorStatus);
         }
         return fn(err);
       }
@@ -367,7 +367,7 @@ HttpContext.prototype.setReturnArgByName = function(name, value) {
         res.status(value);
         break;
       case 'header':
-        res.set(returnDesc.header || name, value);
+        res.set(returnDesc.http.header || name, value);
         break;
     }
   } else {
--- a/lib/http-context.js
+++ b/lib/http-context.js
@@ -38,6 +38,7 @@ function HttpContext(req, res, method, options) {
   this.args = this.buildArgs(method);
   this.methodString = method.stringName;
   this.supportedTypes = this.options.supportedTypes || DEFAULT_SUPPORTED_TYPES;
+  this.result = {};
 
   if (this.supportedTypes === DEFAULT_SUPPORTED_TYPES && !this.options.xml) {
     // Disable all XML-based types by default
@@ -304,6 +305,9 @@ HttpContext.prototype.invoke = function(scope, method, fn, isCtor) {
   var pipe = http && http.pipe;
   var pipeDest = pipe && pipe.dest;
   var pipeSrc = pipe && pipe.source;
+  var ctx = this;
+  var defaultErrorStatus = http && http.errorStatus;
+  var res = this.res;
 
   if (pipeDest) {
     // only support response for now
@@ -313,7 +317,7 @@ HttpContext.prototype.invoke = function(scope, method, fn, isCtor) {
         this.res.header('Content-Type', 'application/json');
         this.res.header('Transfer-Encoding', 'chunked');
 
-        var stream = method.invoke(scope, args, this.options, fn);
+        var stream = method.invoke(scope, args, this.options, ctx, fn);
         stream.pipe(this.res);
         break;
       default:
@@ -324,7 +328,7 @@ HttpContext.prototype.invoke = function(scope, method, fn, isCtor) {
     // only support request for now
     switch (pipeDest) {
       case 'req':
-        this.req.pipe(method.invoke(scope, args, this.options, fn));
+        this.req.pipe(method.invoke(scope, args, this.options, ctx, fn));
         break;
       default:
         fn(new Error('unsupported pipe source'));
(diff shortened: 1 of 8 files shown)


ritch:
test please

lib/rest-adapter.js
```diff
@@ -359,7 +359,9 @@ RestAdapter.errorHandler = function(options) {
         err.status = err.statusCode = 500;
       }
 
-      res.statusCode = err.statusCode || err.status || 500;
+      if (res.statusCode === undefined || res.statusCode === 200) {
```
bajtos: How about other 2xx status codes, like 204 No Content? I am proposing to rewrite the condition to `res.statusCode === undefined || res.statusCode < 400`

lib/http-context.js
```diff
@@ -332,7 +336,41 @@ HttpContext.prototype.invoke = function(scope, method, fn, isCtor) {
...
+  if (returnDesc.root) {
+    this.result = value;
+  } else if (returnDesc.http) {
+    switch (returnDesc.http.target) {
+      case 'status':
+        res.status(value);
+        break;
+      case 'header':
+        res.set(returnDesc.header || name, value);
+        break;
```
bajtos: Could you please add a unit-test for this?

bajtos:
I am not sure if it's a good idea to add a constant `errorStatus` code, I am concerned that it may mask useful status codes like 422 (validation failed), 400 (bad request, caused by a missing argument) etc. Could you please add a unit- test to verify that `err.statusCode` takes precedence over `errorStatus` defined in remoting metadata?

```js
       var method = givenSharedStaticMethod(
         function fn(cb) {
           var err = new Error('test error');
           err.statusCode = 400;
           cb(err);
         },
         {
           http: { status: 201, errorStatus: 508 }
         }
       );
       json(method.url)
         .expect(400, done);
```

bajtos:
Awesome stuff!

It seems to me that passing `ctx` to `SharedMethod.prototype.invoke` is a bit inelegant and creates coupling that may make future refactoring more difficult. I don't have a better solution at hand though, so don't worry about it too much.

ritch:
> I am not sure if it's a good idea to add a constant errorStatus code, I am concerned that it may mask useful status codes like 422 (validation failed), 400 (bad request, caused by a missing argument) etc. 

Well we need some way of declaring the status code for a specific result (eg. when the result is ok send this status, when the result is an error send this status).

I think the `err.status` should take precedence. So if that is set we will just ignore the `errorStatus` default.

crandmck:
@ritch Remote methods' accepts & returns are documented here: http://apidocs.strongloop.com/strong-remoting/#sharedmethod (in source file https://github.com/strongloop/strong-remoting/blob/master/lib/shared-method.js). It might be best for you to add docs, since I'm not familiar with this PR.  But if you want me to do it, please just open a doc issue.

lib/http-context.js
```diff
@@ -384,6 +424,11 @@ HttpContext.prototype.done = function() {
   var data = this.result;
   var res = this.res;
   var accepts = this.req.accepts(this.supportedTypes);
+  var defaultStatus = this.method.http.status;
+
+  if (defaultStatus) {
+    res.status(defaultStatus);
```
bajtos: Should we check that `res.statusCode === undefined || res.statusCode === 200` before applying `defaultStatus`?

bajtos:
The code LGTM, although I left one comment/question for your consideration. Please make sure the CI builds are green before landing this patch.

lib/http-context.js
ritch: Added the check...

auguster:
I'm having a use case with long video processing, sometimes the video is ready (status code 200) or processing (202). You can get the detailed status on a specific method that returns a status 201 if the processing has finished with the URI on where to get it.

So far I'm setting the status code through `new Error(...).statusCode` the problem is that 2xx are not errors and strong-error-handler is replacing status code < 400 by 500.

What has been proposed so far (setting static return status) doesn't work in my use case as I can return 200 or 202.

I'm looking for the best way to return non-error status codes. I guess I could go into the ctx.res and manually set this but that feels so wrong. What should I do ?

bajtos:
@auguster Your method should have a callback argument that's annotated with `http: {target: 'status'}`. See https://loopback.io/doc/en/lb3/Remote-methods.html#argument-descriptions
 and the test-cases checking this functionality:

https://github.com/strongloop/strong-remoting/blob/4ef9baac55a5c35b5b6f8f2252bf0699f758e837/test/rest.test.js#L2052-L2095

auguster:
@bajtos I will try that, thanks.

PR #219, opened 2015-06-26
Fixes the test/streams.js test

jorrit:
Broken since f01b8a6ba0b157f3e6562d11158bc3050a75d61a

### DIFF
--- a/test/streams.js
+++ b/test/streams.js
@@ -29,11 +29,15 @@ describe('strong-remoting', function() {
   }
 
   it('should stream the file output', function(done) {
+    objects.convert('streamopts', function(val) {
+      return JSON.parse(val);
+    });
+
     remotes.fs = fs;
     fs.createReadStream.shared = true;
     fs.createReadStream.accepts = [
       { arg: 'path', type: 'string' },
-      { arg: 'encoding', type: 'string' }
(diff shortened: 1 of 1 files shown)


slnode:

jorrit:
The patch is not yet finished, I need to add uri encoding for the pameters.

jorrit:
It is finished, I squashed the two commits.

jorrit:
The waiting label can be removed.

rmg:
ok to test

bajtos:
@ritch you were working on the stream support in strong-remoting recently, thus I think you are the best person to review this patch?

bajtos:
@jorrit sorry for the long delay in our review process. Are you still keen to get this patch finished?

bajtos:
Actually, I am quite puzzled by this. AFAICT, all tests are passing on the current master, which makes me think this patch is no longer relevant.

@jorrit @ritch thoughts?

jorrit:
I think the node api has changed. At the time I created the PR, node 0.12 just changed the api of fs.createReadStream so the second argument was  an object instead of a string. I recall that this caused the problem that led me to create the PR. However, when looking at the docs of the current node version, it seems that both a string and an object are allowed. So this PR is no longer needed.

PR #221, opened 2015-06-26
HttpContext: do not lowercase type when resolving dynamic converter

jorrit:
Fixes strongloop/loopback#906

### DIFF
--- a/test/http-context.test.js
+++ b/test/http-context.test.js
@@ -1,6 +1,7 @@
 var request = require('supertest');
 var HttpContext = require('../lib/http-context');
 var SharedMethod = require('../lib/shared-method');
+var Dynamic = require('../lib/dynamic');
 var expect = require('chai').expect;
 
 describe('HttpContext', function() {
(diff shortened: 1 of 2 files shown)


slnode:

test/http-context.test.js
```diff
@@ -74,6 +75,24 @@ describe('HttpContext', function() {
...
+      });
+
+      it('should coerce dynamic type with string prop into object', givenMethodExpectArg({
+        type: 'CustomType',
+        input: JSON.stringify({ stringProp: 'string' }),
+        expectedValue: { stringProp: 'string' }
+      }));
+
+      it('should coerce dynamic type with int prop into object', givenMethodExpectArg({
+        type: 'CustomType',
```
ritch: Add a test that breaks before this patch... eg `type: 'customType'`

ritch:
Other than a missing tests that fails before this patch this LGTM

test/http-context.test.js
jorrit: I believe this test will break before the patch. There is no functional difference between `CustomType` and `customType` in this regard, both will fail. Only `customtype` will not fail at the moment.

jorrit:
Can you explain what is wrong with the tests that I added? They will fail when the code change is not there.

PR #227, opened 2015-07-21
fix long lines in docs that were failing jshint

rmg:
Introduced in 4d96596

@crandmck I tried to only wrap the long lines, but I had to take some liberties with the one particularly long URL (including replacing it with a Confluence generated short URL), so you are probably the best to review.

/cc @ritch @raymondfeng because linting is part of this module's tests, doc changes can break the build. For that reason I recommend doc changes go through the PR process so that they can at least be linted before merging.


raymondfeng:
Lgtm

Issue #229, opened 2015-07-27
Method not invoked from file based html application


ritch:
<img width="1268" alt="screen shot 2015-07-30 at 2 39 13 pm" src="https://cloud.githubusercontent.com/assets/462228/8995899/49ec3f44-36c9-11e5-8ee0-dfb3a9c43808.png">
It looks like no HTTP requests are made when running within a file based app (eg. cordova).

OpenSystemsConsulting:
Ping

PR #230, opened 2015-07-30
Explicitly set req.protocol

ritch:
/to @raymondfeng 

This fixes https://github.com/strongloop/strong-remoting/issues/229

Without setting the `req.protocol` the browserify implementation of `http` incorrectly falls back to the `window.location.protocol` (which is `file:`).


raymondfeng:
LGTM

PR #233, opened 2015-08-05
verify the accepts size <= 5000

yorkie:
We should be careful with the `raw` size to call traverse, here because this use case is edge one, so I think the inspect call might have too much consumes on normal remote routing.

R= @ritch 
/cc @raymondfeng

### DIFF
--- a/lib/shared-method.js
+++ b/lib/shared-method.js
@@ -472,7 +472,8 @@ SharedMethod.convertArg = function(accept, raw) {
     )) {
     return raw;
   }
-  if (raw === null || typeof raw !== 'object') {
+  if (raw === null || typeof raw !== 'object' ||
+     util.inspect(raw).length > 5000) {
     return raw;
   }
   var data = traverse(raw).forEach(function(x) {
--- a/lib/shared-method.js
+++ b/lib/shared-method.js
@@ -472,10 +472,16 @@ SharedMethod.convertArg = function(accept, raw) {
     )) {
     return raw;
   }
-  if (raw === null || typeof raw !== 'object' ||
-     util.inspect(raw).length > 5000) {
+  if (raw === null || typeof raw !== 'object') {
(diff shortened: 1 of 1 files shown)


slnode:

raymondfeng:
@yorkie Please sign the CLA at https://cla.strongloop.com/agreements/strongloop/strong-remoting

lib/shared-method.js
```diff
@@ -472,7 +472,8 @@ SharedMethod.convertArg = function(accept, raw) {
     )) {
     return raw;
   }
-  if (raw === null || typeof raw !== 'object') {
+  if (raw === null || typeof raw !== 'object' ||
+     util.inspect(raw).length > 5000) {
```
raymondfeng: This could be expensive too.

lib/shared-method.js
```diff
@@ -472,7 +472,8 @@ SharedMethod.convertArg = function(accept, raw) {
...
     return raw;
   }
   var data = traverse(raw).forEach(function(x) {
```
raymondfeng: I guess that we probably should bypass certain types to avoid traversing into `req` and `res`. Maybe only recurse into plain objects/arrays?

yorkie: So you wanna disable circular object, right?

lib/shared-method.js
yorkie: Do you mean 5000 or the `util.inspect` call?

yorkie:
Signed and replaced the verification as checking `raw.constructor`.

yorkie:
Ping

bajtos:
Please consider adding a unit-test to verify this fix and prevent regressions in the future.

raymondfeng:
@ritch I would like to get your green light.

ritch:
LGTM

PR #239, opened 2015-09-01
Use strict jayson version to prevent regression

bajtos:
The recently released version jayson@1.2.1 introduced a breaking change
in https://github.com/tedeh/jayson/commit/dc0a13dc4 which is causing
our json-rpc adapter to fail, because we are passing a zero-argument
wrapper function to jayson.

This commit is changing the jayson dependency to strict version 1.2.0, which is the latest
version working for us.

/to @ritch @raymondfeng please review

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "eventemitter2": "^0.4.14",
     "express": "4.x",
     "inflection": "^1.7.1",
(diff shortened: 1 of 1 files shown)


ritch:
LGTM

PR #240, opened 2015-09-01
Generate wrapper function with same params as the original one

raymondfeng:
/to @ritch @bajtos 

See https://github.com/tedeh/jayson/commit/dc0a13dc4e50880478a446eb682a75f36ef27e3d

### DIFF
--- a/lib/jsonrpc-adapter.js
+++ b/lib/jsonrpc-adapter.js
@@ -97,6 +97,20 @@ JsonRpcAdapter.errorHandler = function() {
   };
 };
 
+// A mock wrapper function to help code generation via mockWrapper.toString()
+function mockWrapper(method) {
+  return function(__args__) {
+    var args = Array.prototype.slice.call(arguments);
+    if (method.isStatic) {
+      method.getFunction().apply(method.ctor, args);
+    } else {
+      method.sharedCtor.invoke(method, function(err, instance) {
+        method.getFunction().apply(instance, args);
+      });
+    }
+  };
+}
+
 JsonRpcAdapter.prototype.createHandler = function() {
 
   var root = express.Router();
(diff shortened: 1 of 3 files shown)


lib/jsonrpc-adapter.js
```diff
@@ -148,33 +148,41 @@ JsonRpcAdapter.prototype.createHandler = function() {
...
-          method.getFunction().apply(method.ctor, args);
-        } else {
-          method.sharedCtor.invoke(method, function(err, instance) {
-            method.getFunction().apply(instance, args);
-          });
-        }
-      };
-
-      /*
-      I had to this because jayson uses fn.toString to map the parameters,
```
ritch: maybe mention the commit jayson commit?

lib/jsonrpc-adapter.js
```diff
@@ -148,33 +148,41 @@ JsonRpcAdapter.prototype.createHandler = function() {
...
+        'return function(' + argsNames + ') {\n' +
+        '  var args = Array.prototype.slice.call(arguments);\n' +
+        '  if (method.isStatic) {\n' +
+        '    method.getFunction().apply(method.ctor, args);\n' +
+        '  } else {\n' +
+        '    method.sharedCtor.invoke(method, function(err, instance) {\n' +
+        '      method.getFunction().apply(instance, args);\n' +
+        '    });\n' +
+        '  }\n' +
+        '};';
```
bajtos: Can we use `Function.toString` instead of writing the code in a string?

```js
// mockup
var wrapper = function wrapper() {
  var args = //..
  // ... etc
};
var funcBody = wrapper.toString()
  .replace(/^function wrapper()/, 'function(' + argsNames + '));
var fn = new Function('method', funcBody)(method);
```

bajtos:
One minor comment, the rest LGTM.

lib/jsonrpc-adapter.js
```diff
@@ -148,33 +148,41 @@ JsonRpcAdapter.prototype.createHandler = function() {
...
+          var names = [];
+          for (var i = 0; i < m.length - 1; i++) {
+            names.push('param' + i);
+          }
+          argsNames = names.join(',');
+        }
+      }
+      argsNames = argsNames ? argsNames + ',cb' : 'cb';
+      var funcBody =
+        'return function(' + argsNames + ') {\n' +
```
ritch: This is not so great... It probably makes it difficult or impossible for VMs to optimize. That being said, we do cache the result of this function, and once it is an actual function object, I think it should be treated like other functions.

ritch:
After addressing @bajtos comment, this LGTM.

PR #243, opened 2015-09-16
stringify mockWrapper so we can generate coverage

rmg:
This has to be done because when we run the tests with coverage the
function body gets mangled by instrumentation and the string
replacements fail horribly. The end result is the generated function
ends up not even being valid JavaScript and the tests then fail for
obvious reasons.

### DIFF
--- a/lib/jsonrpc-adapter.js
+++ b/lib/jsonrpc-adapter.js
@@ -97,20 +97,26 @@ JsonRpcAdapter.errorHandler = function() {
   };
 };
 
-// A mock wrapper function to help code generation via mockWrapper.toString()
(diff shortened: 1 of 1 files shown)


ritch:
LGTM

bajtos:
Thanks @rmg for fixing this issue.

I wish we could keep using `toString`, so that the function body can be run through the linter and we can use IDE auto-completion while changing its code, but I understand that's not feasible with our CI setup.

rmg:
@bajtos there are some alternatives that would allow the function to remain a function:
 - extract the function to a separate file and exclude that file from coverage instrumentation
 - disable code coverage for the entire project

PR #249, opened 2015-10-26
Support application/vnd.api+json media type

digitalsadhu:
Adds support for application/vnd.api+json media type.

See discussion here: https://github.com/strongloop/loopback/issues/445#issuecomment-150923129

### DIFF
--- a/lib/http-context.js
+++ b/lib/http-context.js
@@ -560,6 +560,10 @@ HttpContext.prototype.done = function(cb) {
   if (dataExists) {
     switch (accepts) {
       case '*/*':
+      case 'application/vnd.api+json':
+        res.header('Content-Type', 'application/vnd.api+json');
+        res.json(data);
+        break;
       case 'application/json':
       case 'json':
         res.json(data);
--- a/test/rest.test.js
+++ b/test/rest.test.js
@@ -295,6 +295,25 @@ describe('strong-remoting-rest', function() {
         .expect('Content-Type', /xml/)
         .end(done);
     });
+
+    it('should treat application/vnd.api+json accept header correctly', function(done) {
+      objects.options.rest = { supportedTypes: ['application/vnd.api+json'] };
+
+      var method = givenSharedStaticMethod(
+        function(cb) { cb(null, { value: 'value' }); },
+        { returns: { arg: 'result', type: 'object' } }
+      );
+
+      request(app).get(method.url)
+        .set('Accept', 'application/vnd.api+json')
+        .expect(200)
+        .expect('Content-Type', /application\/vnd\.api\+json/)
+        .end(function(err, res) {
+          if (err) return done(err);
+          expect(res.body).to.deep.equal({ result: { value: 'value' }});
+          done();
+        });
+    });
   });
 
   describe('cors', function() {


listepo:
:+1:

digitalsadhu:
@bajtos @ritch any chance you guys can have a look at this? Happy to make chances as necessary.

ritch:
@digitalsadhu this looks good to me. Thanks for the patch :smile: :+1:

ritch:
@slnode test please

ritch:
I think we should mention this in the docs somewhere... Any ideas @digitalsadhu ? (this is not a blocker for merging this BTW).

digitalsadhu:
Ok. Will have a scan of the docs and report back.

digitalsadhu:
@ritch 1 thing I've wondered is whether I should have added application/vnd.api+json to the default supported types or not? If not then we should definitely document that the user will need to do this (most likely in config.json I guess) if they want to use application/vnd.api+json in the accept header and have it recognized. Thoughts?

digitalsadhu:
On that note, adding documentation here https://docs.strongloop.com/display/public/LB/config.json#config.json-Remotingproperties
Would be one place

ritch:
Here are the default types: https://github.com/strongloop/strong-remoting/pull/249/files#diff-c8e4ad547ed92deb50d4eb3bc6430335R18

The issue I see adding as a default is, when a client says they accept this type of JSON, we aren't using that information to give them what they are asking for. So for now we should just add a note to the docs saying that you can opt into the support (eg. what you did in your test).

@crandmck can you make this change ^^^ ?

digitalsadhu:
Yup. Sounds good to me.

crandmck:
Re docs
> can you make this change ^^^ ?

Sure, but I will wait for the PR to be merged, which I assume s/be soon.

crandmck:
Went ahead and added to docs, since I presume the merge will happen soon. Thx for the heads up!

digitalsadhu:
@ritch Whats the process for merging now? Need me to do anything further?

ritch:
@digitalsadhu sorry for delayed responses :( - pinging me via a comment is the best way to bring something like this to the top of my queue. Thanks again for the contribution.

digitalsadhu:
@ritch @crandmck 
Just noticed the doc update here under supportedTypes:
https://docs.strongloop.com/display/public/LB/config.json

I think it isn't quite correct. We decided not to add application/vnd.api+json to the default set of supported types so it needs to be added manually. The docs have it under "default". It's more that it's one of the possible options you can set but not there by default.

crandmck:
@digitalsadhu Thanks for clarifying!  I updated the doc.  
Can  you tell me how one adds it as a supportedType?  That would probably be useful to note as well.

digitalsadhu:
@crandmck great!

You can add it as a supportedType either in config.json as such:
```json
{
  "remoting": {
    "rest": {
      "supportedTypes": ["application/vnd.api+json", "application/json", ...]
    }
  }
}
```

or programatically via something like:
```js
app.remotes().options.rest.supportedTypes =  ['application/vnd.api+json', 'application/json', ...]
```

I think, if at all possible, it would be good to document somewhere that setting the `Accept` header on the request to `application/vnd.api+json` will result in the `Content-Type` header being automatically set to `application/vnd.api+json` on the response so long as `application/vnd.api+json` is in the array of supported types.

crandmck:
Thanks--I added that last part to https://docs.strongloop.com/display/LB/Exposing+models+over+REST#ExposingmodelsoverREST-Requestformat.

digitalsadhu:
Nice, awesome!

Issue #251, opened 2015-11-04
Last Update for Accept "application/vnd.api+json" Breaks Apps[Urgent]


jaime-franco:
Responses now are default "application/vnd.api+json", in the case in the file  lib/http-context.js is the returned when accepts is */*.

https://github.com/strongloop/strong-remoting/commit/d7d9a73cc77f33c13f31d3c8e57af8277e33363c

jaime-franco:
If some one has the same problem you can add in the config file

    "rest": {
      "supportedTypes": [ "application/json"]
    }

But it will be great if the default setting is json instead of vnd.api+json

voitau:
Could someone from Strongloop confirm that the fix suggested by @jaime-franco is a proper fix/workaround? Thanks!

jaime-franco:
Hi, @voitau  i found the config information on this link :

https://docs.strongloop.com/display/public/LB/config.json#config.json-Remotingproperties

And i think this is a issue because in the documentation you can read : 

NOTE: 'application/vnd.api+json' is supported, but is not one of the default types. 

And actually this las update change that.

raymondfeng:
Can you submit a patch so that `case '*/*':` goes with `case 'json':`?

jaime-franco:
Ok

voitau:
Thanks for a quick fix! Is there a release with this fix planned?

PR #252, opened 2015-11-06
Fix issue #251, now default responses are application/json

jaime-franco:


slnode:

raymondfeng:
Thanks for the patch. It would be nice to add a test case to cover `*/*`.

raymondfeng:
@bajtos or @ritch Can you review?

bajtos:
:+1:

PR #256, opened 2015-11-17
Close server stream once event-stream consumer disconnects

kraman:
Connect to strongloop-internal/scrum-nodeops#945

### DIFF
--- a/lib/http-context.js
+++ b/lib/http-context.js
@@ -488,6 +488,16 @@ HttpContext.prototype.respondWithEventStream = function(stream) {
   stream.on('end', function() {
     client.send({event: 'end', data: 'null'});
   });
+
+  function closeStream() {
+    if (stream) {
+      stream.end();
+      stream = null;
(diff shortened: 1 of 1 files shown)


kraman:
Based on https://nodejs.org/api/http.html#http_class_http_serverresponse looks like there is no `end` event on http response objects.

kraman:
@ritch Looks like tests are failing for unrelated reasons.

bajtos:
@kraman could you please add a unit-test to cover this change? I think it's really important to have an automated test in place, considering the uncertainty about the correctness of the implementation.

@piscisaureus could you perhaps take a look and let us know if we are doing this correctly? The idea is to detect when the browser client closes our long-living HTTP response stream, so that we can stop producing data that would be discarded.

kraman:
@ritch pointed out that this is not the correct place to close the stream. It should be closed by the method that creates/manages the stream.

PR #259, opened 2015-12-02
Enable default values for arguments

jsheely:
Loopback explorer utilises a default value property on arguments in order to pre-populate input fields. This default is also exposed in the swagger.json. This simply utilizes that default property when invoking the remote method if a value has not been defined in the request.

### DIFF
--- a/test/shared-method.test.js
+++ b/test/shared-method.test.js
@@ -79,6 +79,30 @@ describe('SharedMethod', function() {
   });
 
   describe('sharedMethod.invoke', function() {
+    it('returns default value on custom remote method when parameter is not passed', function(done){
+      var defaultValue = 100;
+      var method = givenSharedMethod(
+        function(num, next) { 
+          return new Promise(function(resolve, reject) {
+            if(typeof num === 'number'){
+              return resolve({num: num});
+            }
+            return reject(new Error('Parameter is not a number'));
+          });
+        },
+        {
+          accepts: [{ arg: 'num', type: 'number', default:defaultValue }],
+          returns: {root:true, type:'object'}
+      });
+      
+      method.invoke('ctx', {}, function(err, result){
+        setImmediate(function(){
+          expect(result).to.exist;
+          expect(result.num).to.equal(defaultValue);
+          done();
+        });
+      });
+    });
     it('returns 400 when number argument is `NaN`', function(done) {
       var method = givenSharedMethod({
         accepts: [{ arg: 'num', type: 'number' }]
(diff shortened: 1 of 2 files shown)


slnode:

bajtos:
@slnode test please

bajtos:
> Loopback explorer utilises a default value property on arguments in order to pre-populate input fields. This default is also exposed in the swagger.json. This simply utilizes that default property when invoking the remote method if a value has not been defined in the request.

@jsheely thank you for the pull request, please add a unit-test to verify your implementation and prevent regressions in the future.

@ritch @raymondfeng @fabien @STRML I feel this change may cause subtle side effects. Do you happen to know why strong-remoting was not honouring `default` setting in the first place? Can you think of any use case where this change would break existing applications?

jsheely:
Trying to add a unit-test to the project. However when returning a promise resolve() it never appears to invoke done and then will not call the callback to test the result.
```
Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.
````

I get the same result on all the promise based test methods. However if I reject the promise it works calls the callback with the error without issue.

What am I missing?

**Comment by jsheely** [comment]
Nevermind. I think this was a result of another update I was making. Disregard

**Comment by 0candy** [comment]
Thanks @jsheely for the unit test. Could you please fix up the trailing whitespace to pass eslint? You can run `grunt eslint` to see which lines. Thanks!

**Comment by slnode** [comment]

File: test/shared-method.test.js [hunk 1]
```diff
@@ -79,6 +79,30 @@ describe('SharedMethod', function() {
   });
 
   describe('sharedMethod.invoke', function() {
+    it('returns default value on custom remote method when parameter is not passed', function(done){
+      var defaultValue = 100;
+      var method = givenSharedMethod(
+        function(num, next) { 
+          return new Promise(function(resolve, reject) {
```
bajtos: I think this can test be simplified a bit to something line the following:

```js
function(num) {
  if (typeof num !== 'number') {
    return Promise.reject(new Error(...));
  }
  return Promise.resolve({ num: num });
}
```

test/shared-method.test.js
```diff
@@ -79,6 +79,30 @@ describe('SharedMethod', function() {
...
+            return reject(new Error('Parameter is not a number'));
+          });
+        },
+        {
+          accepts: [{ arg: 'num', type: 'number', default:defaultValue }],
+          returns: {root:true, type:'object'}
+      });
+      
+      method.invoke('ctx', {}, function(err, result){
+        setImmediate(function(){
```
bajtos: Why is `setImmediate` needed?

bajtos:
@jsheely sorry for the long delay in our review process :( Are you still keen to get this finished? If so, then please rebase the patch on top of the current master and make sure `npm test` passes on your machine.

bajtos:
I am closing this pull request as abandoned. Feel free to reopen when you get time to address the comments above.

PR #260, opened 2015-12-04
Don't coerce null to 'null' with string accepts in shared-method.

STRML:
A method definition with a string argument would coerce `null` to `'null'`. 

connect to #263


STRML:
Build doesn't fail here, and I can't check the status because your CI is locked down. Status?

bajtos:
@slnode test please

bajtos:
Closing in favour of https://github.com/strongloop/strong-remoting/pull/349. `null` values are rejected for string arguments now.

PR #261, opened 2015-12-04
Fix incorrect boolean logic on shared-method's `documented` flag.

STRML:
Boolean comparisons don't work like that - as written it's not possible to actually make the `documented` flag stick.

### DIFF
--- a/lib/shared-method.js
+++ b/lib/shared-method.js
@@ -96,7 +96,7 @@ function SharedMethod(fn, name, sc, options) {
   this.description = options.description || fn.description;
   this.accessType = options.accessType || fn.accessType;
   this.notes = options.notes || fn.notes;
-  this.documented = (options.documented || fn.documented) !== false;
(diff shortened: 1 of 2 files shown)


bajtos:
@STRML nice catch. Could you please add a unit-test to verify your implementation and prevent regressions in the future?

STRML:
@bajtos Ready to go

STRML:
What's going on with the CI? This build succeeds, but has been pending/failed for days.

bajtos:
> What's going on with the CI? This build succeeds, but has been pending/failed for days.

When one of the dependant projects fails, our CI stops all other pending jobs. Unfortunately the github status is not updated :(

bajtos:
@STRML The patch LGTM now. Could you please squash the commits to a single one?

STRML:
@bajtos :shipit:

bajtos:
Landed, thank you for the contribution :)

PR #270, opened 2015-12-16
Retain accepted content-type with no-content

davidcheung:
#### rest.test.js
- Add test case to make sure default stays as `application/json`
- Add test case to make sure it retains specified content-type

#### http-context.js
- seperate logic of figuring out what to send and sending it

Connected to strongloop/strong-remoting/issues/258

### DIFF
--- a/lib/http-context.js
+++ b/lib/http-context.js
@@ -491,6 +491,74 @@ HttpContext.prototype.respondWithEventStream = function(stream) {
 };
 
 /**
+ * Utility functions to send response body
+ */
+function sendBodyJson(res, data) {
+  res.json(data);
+}
+
+function sendBodyJsonp(res, data) {
+  res.jsonp(data);
+}
+
+function sendBodyXml(res, data) {
+  if (data === null) {
+    res.header('Content-Length', '7');
+    res.send('<null/>');
+  } else if (data) {
+    try {
+      var xml = toXML(data);
+      res.send(xml);
+    } catch (e) {
+      res.status(500).send(e + '\n' + data);
+    }
+  }
+}
+
+function sendBodyDefault(res) {
+  res.status(406).send('Not Acceptable');
+}
+
+/**
+ * Deciding on the operation of response, function is called inside this.done()
+ */
+
+HttpContext.prototype.resolveReponseOperation = function(accepts) {
+  var result = { // default
+    sendBody : sendBodyJson,
+    contentType : 'application/json'
+  };
+  switch (accepts) {
+    case '*/*':
+    case 'application/json':
+    case 'json':
+      break;
+    case 'application/vnd.api+json':
+      result.contentType = 'application/vnd.api+json';
+      break;
+    case 'application/javascript':
+    case 'text/javascript':
+      result.sendBody = sendBodyJsonp;
+      break;
+    case 'application/xml':
+    case 'text/xml':
+    case 'xml':
+      if (accepts == 'application/xml') {
+        result.contentType = 'application/xml';
+      } else {
+        result.contentType = 'text/xml';
+      }
+      result.sendBody = sendBodyXml;
+      break;
+    default:
+      result.sendBody = sendBodyDefault;
+      result.contentType = 'text/plain';
+      break;
+  }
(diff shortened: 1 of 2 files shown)


davidcheung:
@bajtos I have refactored the implementation, can you take a look to provide some feedback to see if it make sense?

lib/http-context.js
```diff
@@ -491,6 +491,60 @@ HttpContext.prototype.respondWithEventStream = function(stream) {
...
+          } catch (e) {
+            res.status(500).send(e + '\n' + data);
+          }
+        }
+      };
+      break;
+    default:
+      result.operation = function() {
+        res.status(406).send('Not Acceptable');
+      };
```
bajtos: I think we should set content-type to `text/plain` here.

lib/http-context.js
```diff
@@ -556,55 +610,19 @@ HttpContext.prototype.done = function(cb) {
...
-          } catch (e) {
-            res.status(500).send(e + '\n' + data);
-          }
-        }
-        break;
-      default:
-        res.status(406).send('Not Acceptable');
-        break;
-    }
+    res.header('Content-Type', operationResults.contentType);
```
bajtos: You are setting the header twice - once at L614 and second time here.

lib/http-context.js
```diff
@@ -491,6 +491,60 @@ HttpContext.prototype.respondWithEventStream = function(stream) {
...
 
 /**
+ * Deciding on the operation of response, function is called inside this.done()
+ */
+
+HttpContext.prototype.resolveReponseOperation = function(accepts) {
+  var res = this.res;
+  var data = this.result;
+  var result = { // default
+    operation : function() { res.json(data); },
```
bajtos: Perhaps `sendBody` would be a better name than `operation`?

test/rest.test.js
```diff
@@ -2046,6 +2046,33 @@ describe('strong-remoting-rest', function() {
       .end(done);
   });
 
+  it('retain content-type with no-content', function(done) {
```
bajtos: please mention "xml" in the test name

test/rest.test.js
```diff
@@ -2046,6 +2046,33 @@ describe('strong-remoting-rest', function() {
...
+
+    request(app)
+      .get(method.url + '?arg={"x":1}&_format=xml')
+      .expect(204)
+      .end(function(err, res) {
+        expect(res.get('Content-type')).to.match(/xml/);
+        done();
+      });
+  });
+  it('default content-type to be application/json', function(done) {
```
bajtos: add an empty line to visually delimit test cases.

bajtos: Also change the test name to read like a sentence please. E.g. `it('defaults content-type to application/json')`

test/rest.test.js
bajtos: e.g. `it('returns correct content-type in an empty XML response')`

lib/http-context.js
bajtos: Performance micro-optimization, hopefully not preliminary:

At the moment, a new function must be created for each HttpContext instance because the closure is always different. IIRC, this prevents V8 from optimising the code. Please rework this function to get `res` and `data` from arguments, not the closure, i.e. `function(res, data) { res.json(data); }`. That should allow V8 to optimise it more.

Even better, define these small functions outside of `HttpContext.prototype.resolveResponseOperation`.

```js
HttpContext.prototype.resolveReponseOperation = function(accepts) {
  // ...
  result.sendBody = sendJson;
  // ...
}

function sendJson(res, data) {
  res.json(data);
}
```

bajtos:
@davidcheung cool, the new implementation looks great! I left few comments on what to improve, this patch is almost good to be landed.

Please ping @ritch to see if he has time to give this the final review and land it while I am offline. If he hasn't, then I can land it on Monday.

davidcheung:
@bajtos I have updated the code as per suggestions, looks much nicer now. 

@bajtos or @ritch, please review

bajtos:
LGTM. Landed, thank you!

PR #271, opened 2015-12-28
Allow source to be `formData` (=== `form`) to comply with swagger-definition

0ff:
Instead of changing the generated output (which would seem to be the better choice), I suggest allowing this as an alias. The reason is that both the [swagger generator](https://github.com/strongloop/loopback-swagger/blob/9a30a143a47445f321c1d349c5b808517e248686/lib/codegen/spec-converter.js#L81) and the swagger ui use `formData` instead of just `form`. In my opinion there is no standard that said "use `form`" and thus it was chosen at will. Migrating to `formData` would bring loopback closer to the swagger specs.

### DIFF
--- a/lib/http-context.js
+++ b/lib/http-context.js
@@ -136,6 +136,7 @@ HttpContext.prototype.buildArgs = function(method) {
               val = ctx.req.body;
               break;
             case 'form':
+            case 'formData':
               // From the form (body)
               val = ctx.req.body && ctx.req.body[name];
               break;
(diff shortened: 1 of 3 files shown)


slnode:

bajtos:
I have fixed the commit message to follow the [50/72 rule](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html) and landed the change via a1b156b (master/3.0) and b7b9dc66 (2.x)

bajtos:
Thank you for the contribution!

PR #272, opened 2016-01-06
To allow user to customize the root element of xml output

### LINKED ISSUE #262: Customize the name of the root XML element
At the moment, when producing a response in the XML format, strong-remoting always wrap the response body in `<response>` element ([source code](https://github.com/strongloop/strong-remoting/blob/32efdc433f8bd12f1ecc7872c855216b99c0911e/lib/http-context.js#L448-L457)).

We should allow developers to customise the name of this root element, for example to produce a response body like this:

```xml
<foo>bar</foo>
```

A possible solution is to implement a new remoting setting at the shared-method level:

```js
var fn = function(cb) {
  cb(null, 'bar');
};

setupRemoting(fn, {
  "accepts": [],
  "returns": { "type": "string", "root": true },
  "xml": { "rootElement": "myCustomName" } }
});
```

davidcheung:
To take wrapperElement as an option in remoteMethod setup 
```yaml
{
  returns: { 
    root: true,
      xml: {wrapperElement: 'a'}
    },
    http: { path: '/' }
}
```

Objective is to allow user to produce output as follow 
```xml
<a>1</a>
```

Fixes #262

connected to strongloop/strong-remoting/issues/262

### DIFF
--- a/lib/http-context.js
+++ b/lib/http-context.js
@@ -433,19 +433,26 @@ function toJSON(input) {
   }
 }
 
-function toXML(input) {
+function toXML(input, options) {
   var xml;
+  var xmlDefaultOptions = { declaration : true };
+  var xmlOptions = util._extend(xmlDefaultOptions, options);
   if (input && typeof input.toXML === 'function') {
     xml = input.toXML();
   } else {
-    if (input) {
+    if (typeof input == 'string') {
+      //  this is a hack to handle js2xmlparser throws syntax error on string
+      //  wraps string from "a string" to "\"a string\""
+      input = JSON.stringify(input);
+    } else if (typeof input == 'object') {
       // Trigger toJSON() conversions
       input = toJSON(input);
     }
     if (Array.isArray(input)) {
       input = { result: input };
     }
-    xml = js2xmlparser('response', input, {
+    xml = js2xmlparser(xmlOptions.wrapperElement || 'response', input, {
+      declaration : { include : xmlOptions.declaration },
       prettyPrinting: {
         indentString: '  '
       },
@@ -501,13 +508,14 @@ function sendBodyJsonp(res, data) {
   res.jsonp(data);
 }
 
-function sendBodyXml(res, data) {
+function sendBodyXml(res, data, method) {
   if (data === null) {
     res.header('Content-Length', '7');
     res.send('<null/>');
   } else if (data) {
     try {
-      var xml = toXML(data);
+      var xmlOptions = method.returns[0].xml || {};
+      var xml = toXML(data, xmlOptions);
       res.send(xml);
     } catch (e) {
       res.status(500).send(e + '\n' + data);
@@ -629,7 +637,7 @@ HttpContext.prototype.done = function(cb) {
     res.header('Content-Type', operationResults.contentType);
   }
   if (dataExists) {
-    operationResults.sendBody(res, data);
+    operationResults.sendBody(res, data, method);
   } else {
     if (res.statusCode === undefined || res.statusCode === 200) {
       res.statusCode = 204;
--- a/test/rest.test.js
+++ b/test/rest.test.js
@@ -1359,6 +1359,121 @@ describe('strong-remoting-rest', function() {
           });
       });
 
+      it('should allow customized xml root element', function(done) {
+        var method = givenSharedStaticMethod(
+          function bar(cb) {
+            cb(null, {a: 1, b: 2});
+          },
+          {
+            returns: {
+              arg: 'data', type: 'object', root: true,
+              xml: { wrapperElement: 'foo'}
+            },
+            http: { path: '/' }
+          }
+        );
+        request(app).get(method.classUrl)
+          .set('Accept', 'application/xml')
+          .set('Content-Type', 'application/json')
+          .send()
+          .expect('Content-Type', /xml/)
+          .expect(200, function(err, res) {
+            expect(res.text).to.equal(
+              '<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n' +
+              '<foo>\n  ' +
+                '<a>1</a>\n  ' +
+                '<b>2</b>\n' +
+              '</foo>');
+            done(err, res);
+          });
+      });
+
+      it('should allow xml declaration to be disabled', function(done) {
+        var method = givenSharedStaticMethod(
+          function bar(cb) {
+            cb(null, {a: 1, b: 2});
+          },
+          {
+            returns: {
+              arg: 'data', type: 'object', root: true,
+              xml: { declaration : false }
+            },
+            http: { path: '/' }
+          }
+        );
+        request(app).get(method.classUrl)
+          .set('Accept', 'application/xml')
+          .set('Content-Type', 'application/json')
+          .send()
+          .expect('Content-Type', /xml/)
+          .expect(200, function(err, res) {
+            expect(res.text).to.equal(
+              '<response>\n  ' +
+                '<a>1</a>\n  ' +
+                '<b>2</b>\n' +
+              '</response>');
+            done(err, res);
+          });
+      });
+
+      it('should allow string results to output as xml', function(done) {
+        var method = givenSharedStaticMethod(
+          function bar(cb) {
+            var stringResult = 'a quick brown fox jumps over the lazy dog';
+            cb(null, stringResult);
+          },
+          {
+            returns: {
+              root: true,
+              xml: { wrapperElement : 'text' }
+            },
+            http: { path: '/' }
+          }
+        );
+        request(app).get(method.classUrl)
+          .set('Accept', 'application/xml')
+          .set('Content-Type', 'application/json')
+          .send()
+          .expect('Content-Type', /xml/)
+          .expect(200, function(err, res) {
+            expect(res.text).to.equal(
+              '<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n' +
+              '<text>a quick brown fox jumps over the lazy dog' +
+              '</text>');
+            done(err, res);
+          });
+      });
+
+      it('should handle UTF-8 & special & reserved characters', function(done) {
+        var method = givenSharedStaticMethod(
+          function bar(cb) {
+            var stringA = 'foo\xC1\xE1\u0102\u03A9asd><=$~!@#$%^&*()-_=+/.,;\'"[]{}?';
+            cb(null, {a: stringA});
+          },
+          {
+            returns: {
+              arg: 'data', type: 'object', root: true,
+              xml: { wrapperElement: false }
+            },
+            http: { path: '/' }
+          }
+        );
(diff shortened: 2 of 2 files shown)


lib/http-context.js
```diff
@@ -433,19 +433,28 @@ function toJSON(input) {
...
+  } else {
+    wrapperElement = 'response';
+  }
   if (input && typeof input.toXML === 'function') {
     xml = input.toXML();
   } else {
-    if (input) {
+    if (typeof(input) == 'string') {
+      input = { result: input };
+    } else if (typeof(input) == 'object') {
```
davidcheung: @bajtos when `input` is string, `js2xmlparser` gets syntax error, then try catch block from `sendBodyXml()` throws 500 , should we modify the data so it doesn't internal server error? or should we not let user pass in just one string.

lib/http-context.js
```diff
@@ -456,6 +465,11 @@ function toXML(input) {
       }
     });
   }
+  if (returns.xml && returns.xml.wrapperElement === false) {
+    //allowing XML output to have no root element
+    var match = xml.match(/<response>([\s\S]*?)<\/response>$/);
+    xml = match[1].replace(/^[\s]{0,2}/gim, '').trim();
+  }
```
davidcheung: @bajtos the current plugin we are using to parse javascript to XML does not allow JSON objects with no [`root element`](https://github.com/michaelkourlas/node-js2xmlparser#usage), do you have any suggestion on how to approach this? Using a different plugin will cause the way of attributes handling to be different.

lib/http-context.js
```diff
@@ -433,19 +433,25 @@ function toJSON(input) {
...
+function toXML(input, options) {
   var xml;
+  var xmlDefaultOptions = { declaration : true };
+  var xmlOptions = util._extend(xmlDefaultOptions, options);
   if (input && typeof input.toXML === 'function') {
     xml = input.toXML();
   } else {
-    if (input) {
+    if (typeof(input) == 'string') {
+      // wraps string from "a string" to "\"a string\""
```
davidcheung: explaining why we need to do that

test/rest.test.js
```diff
@@ -1359,6 +1359,58 @@ describe('strong-remoting-rest', function() {
...
+          .expect('Content-Type', /xml/)
+          .expect(200, function(err, res) {
+            console.log(res.text);
+            expect(res.text).to.equal(
+              '<a>1</a>\n' +
+              '<b>2</b>');
+            done(err, res);
+          });
+      });
+
```
davidcheung: add test to make sure UTF8 is supported

lib/http-context.js
```diff
@@ -456,6 +462,11 @@ function toXML(input) {
       }
     });
   }
+  if (xmlOptions.wrapperElement === false) {
+    //allowing XML output to have no root element
+    var match = xml.match(/<response>([\s\S]*?)<\/response>$/);
+    xml = match[1].replace(/^[\s]{0,2}/gim, '').trim();
+  }
```
davidcheung: removing this because we probably shouldn't support `non-xml` outputs while sending the `XML header`

davidcheung:
also added `declaration=false` flag 
```json
{
"returns": { 
  "root": true,
    "xml": { declaration: false }
  },
  "http": { "path": "/" }
}
```
to not include the declaration of
```xml
<?xml version="1.0" encoding="UTF-8"?>
```

To address this https://github.com/strongloop-internal/scrum-cs/issues/396
>Everytime I return something like:
```
cb(null, '{a:blah}');
```
I receive xml like: `<response><a>blah</a></response>`
But what I want is xml like: `<a>blah</a>`

@bajtos, @ritch not sure if this feature is needed, and not sure where to add documentation for such, any advice?

ritch:
> To address this strongloop-internal/scrum-cs#396

I don't think this exactly addresses the issue, but it might be a welcome feature.

lib/http-context.js
```diff
@@ -433,19 +433,26 @@ function toJSON(input) {
...
-function toXML(input) {
+function toXML(input, options) {
   var xml;
+  var xmlDefaultOptions = { declaration : true };
+  var xmlOptions = util._extend(xmlDefaultOptions, options);
   if (input && typeof input.toXML === 'function') {
     xml = input.toXML();
   } else {
-    if (input) {
+    if (typeof(input) == 'string') {
```
ritch: use `typeof input` not `typeof(input)`

lib/http-context.js
```diff
@@ -433,19 +433,26 @@ function toJSON(input) {
...
+      //  this is a hack to handle js2xmlparser throws syntax error on string
+      //  wraps string from "a string" to "\"a string\""
+      input = JSON.stringify(input);
```
ritch: this could be

```js
input = '"' + input + '"';
```

couldn't it?

davidcheung: @ritch  we could, but we will also need to handle escaping the double quotes in the string itself, not sure any other stuff `JSON.stringify` escapes, I feel JSON.stringify might be easier to maintain, any thoughts?

davidcheung:
@ritch, I have addressed the 2 suggestions you made, please review

davidcheung:
@bajtos review please :)

davidcheung:
@slnode test please

lib/http-context.js
```diff
@@ -433,19 +433,26 @@ function toJSON(input) {
   }
 }
 
-function toXML(input) {
+function toXML(input, options) {
   var xml;
+  var xmlDefaultOptions = { declaration : true };
```
bajtos: AFAICT, this is adding another feature, which is to allow the developer to specify whether XML declaration is included in the response. I can live with the fact that it's introduced together with other changes, but please add a unit-test to verify.

lib/http-context.js
```diff
@@ -433,19 +433,26 @@ function toJSON(input) {
...
+  var xmlOptions = util._extend(xmlDefaultOptions, options);
   if (input && typeof input.toXML === 'function') {
     xml = input.toXML();
   } else {
-    if (input) {
+    if (typeof input == 'string') {
+      //  this is a hack to handle js2xmlparser throws syntax error on string
+      //  wraps string from "a string" to "\"a string\""
+      input = JSON.stringify(input);
```
bajtos: Please add a unit-test to verify this case.

test/rest.test.js
```diff
@@ -1359,6 +1359,61 @@ describe('strong-remoting-rest', function() {
...
+              '<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n' +
+              '<foo>\n  ' +
+                '<a>1</a>\n  ' +
+                '<b>2</b>\n' +
+              '</foo>');
+            done(err, res);
+          });
+      });
+
+      it('should handle UTF-8 characters properly', function(done) {
```
bajtos: This test is testing not only UTF-8, but also special/reserved characters. Please update the test name to reflect that.

bajtos:
@davidcheung reviewed, see my comments above.

davidcheung:
hi @bajtos, have added unit test for the changes. Please review again :)

test/rest.test.js
```diff
@@ -1359,6 +1359,112 @@ describe('strong-remoting-rest', function() {
...
+        );
+        request(app).get(method.classUrl)
+          .set('Accept', 'application/xml')
+          .set('Content-Type', 'application/json')
+          .send()
+          .expect('Content-Type', /xml/)
+          .expect(200, function(err, res) {
+            expect(res.text).to.equal(
+              '<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n' +
+              '<response>a quick brown fox jumps over the lazy dog' +
```
bajtos: Please modify the test to use a custom root element, e.g. `<text>` instead of `<response>`.

test/rest.test.js
```diff
@@ -1359,6 +1359,112 @@ describe('strong-remoting-rest', function() {
...
       });
 
+      it('should allow customized xml root element', function(done) {
+        var method = givenSharedStaticMethod(
+          function bar(cb) {
+            cb(null, {a: 1, b: 2});
+          },
+          {
+            returns: { arg: 'data', type: 'object', root: true,
+            xml: { wrapperElement: 'foo'} },
```
bajtos: please reindent these two lines to make it clear that `xml` is nested under `returns`:

```js
returns: {
  arg: ... root: true,
  xml: {...}
},
http: {...}
```

test/rest.test.js
```diff
@@ -1359,6 +1359,112 @@ describe('strong-remoting-rest', function() {
...
+      });
+
+      it('should allow xml decorations to be disabled', function(done) {
+        var method = givenSharedStaticMethod(
+          function bar(cb) {
+            cb(null, {a: 1, b: 2});
+          },
+          {
+            returns: { arg: 'data', type: 'object', root: true,
+            xml: { declaration : false } },
```
bajtos: Same comment here - please reindent.

test/rest.test.js
```diff
@@ -1359,6 +1359,112 @@ describe('strong-remoting-rest', function() {
...
+
+      it('should handle UTF-8 & special & reserved characters', function(done) {
+        var method = givenSharedStaticMethod(
+          function bar(cb) {
+            var stringA = 'foo\xC1\xE1\u0102\u03A9asd><=$~!@#$%^&*()-_=+/.,;\'"[]{}?';
+            cb(null, {a: stringA});
+          },
+          {
+            returns: { arg: 'data', type: 'object', root: true,
+            xml: { wrapperElement: false } },
```
bajtos: And here too.

bajtos:
@davidcheung great, this is almost ready. Please address my comments above and squash the commits into a single one. Make sure the commit message mentions all new features (`declaration: false`, `wrapperElement: 'name'`).

lib/http-context.js
davidcheung: Unit-test added.

lib/http-context.js
davidcheung: Unit-test added.

davidcheung:
@bajtos i have addressed the comments above, I believe the PR is good to go. Please review again

test/rest.test.js
```diff
@@ -1359,6 +1359,121 @@ describe('strong-remoting-rest', function() {
...
+              '<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n' +
+              '<foo>\n  ' +
+                '<a>1</a>\n  ' +
+                '<b>2</b>\n' +
+              '</foo>');
+            done(err, res);
+          });
+      });
+
+      it('should allow xml decorations to be disabled', function(done) {
```
bajtos: `...xml declaration...`, not decoration.

bajtos:
@slnode test please

bajtos:
@davidcheung I have fixed the test name myself and improved the commit message a bit. The patch LGTM, I'll land it as soon as I get CI build results.

PR #273, opened 2016-01-13
Start development of 3.0

0candy:
* Update version number in package.json, publish under "next" tag
* Add 3.0-RELEASE-NOTES.md to incrementally build release docs

Connect to https://github.com/strongloop-internal/scrum-loopback/issues/617

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,10 @@
     "Remoting",
     "REST"
   ],
-  "version": "2.24.0",
+  "version": "3.0.0-alpha.1",
+  "publishConfig": {
+    "tag": "next"
+  },
   "scripts": {
     "pretest": "grunt jscs jshint",
     "test": "mocha"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -7,7 +7,8 @@
     "Remoting",
     "REST"
   ],
-  "version": "2.24.0",
+  "version": "3.0.0-alpha.1",
+  "publishConfig": { "tag": "next" },
```
superkhau: `"publishConfig": {"tag": "next"},`

or copy style from loopback: https://github.com/strongloop/loopback/blob/master/package.json#L3-L6

bajtos: In LoopBack, we actually use spaces inside curly brackets, as that's the style used in Node core. `{ "tag": "next" }`.

However, in this particular case, it's better to put the contents on new line, because that's what npm will do the next time somebody runs `npm install --save`.

```json
{
  "publishConfig": {
    "tag": "next"
  }
}
```

bajtos:
@0candy one minor comment above, otherwise this should be good to land.

bajtos:
LGTM :+1:

PR #274, opened 2016-01-20
Implement custom API router

gunjpan:
Initial commmit


gunjpan:
closed due to incorrect commit!

PR #275, opened 2016-01-25
Change method call to find remote method.

0candy:
Connect to https://github.com/strongloop-internal/scrum-loopback/issues/617

See also https://github.com/strongloop/loopback/issues/597

### DIFF
--- a/3.0-RELEASE-NOTES.md
+++ b/3.0-RELEASE-NOTES.md
@@ -5,3 +5,25 @@ always describe the impact on users and instructions for upgrading
 applications from 2.x to 3.0.
 
 See also https://github.com/strongloop/loopback/blob/master/3.0-DEVELOPING.md
+
+
+## Invocation for remote method has been changed.
+
+We are deprecating the `SharedClass.prototype.find` and
+`SharedClass.prototype.disableMethod` for looking up and disabling remote
+methods in 3.0. They will be replaced by `SharedClass.prototype.findMethodByName`
+and `SharedClass.prototype.disableMethodByName` where you can pass in a static
+method or a prototype method name. These new methods will accept a string in the
+form of "name" for a static method and "prototype.name" for a prototype method.
+
+To find a static remote method:
+`findMethodByName('create')`
+
+To find a prototype remote method:
+`findMethodByName('prototype.updateAttributes')`
+
+To disable a static remote method
+`disableMethodByName('create')`
+
+To disable a prototype remote method:
+`disableMethodByName('prototype.updateAttributes')`
--- a/lib/shared-class.js
+++ b/lib/shared-class.js
@@ -1,6 +1,7 @@
 /*!
  * Expose `SharedClass`.
  */
+var deprecated = require('depd')('strong-remoting');
 
 module.exports = SharedClass;
 
@@ -200,6 +201,8 @@ SharedClass.prototype.resolve = function(resolver) {
  */
 
 SharedClass.prototype.find = function(fn, isStatic) {
+  deprecated('SharedClass.prototype.find is deprecated. ' +
+    'Use SharedClass.prototype.findMethodByName instead.');
   var methods = this.methods();
   return find(methods, fn, isStatic);
 };
@@ -212,9 +215,35 @@ SharedClass.prototype.find = function(fn, isStatic) {
  */
 
 SharedClass.prototype.disableMethod = function(fn, isStatic) {
-  var disableMethods = this._disabledMethods;
+  deprecated('SharedClass.prototype.disableMethod is deprecated. ' +
+    'Use SharedClass.prototype.disableMethodByName instead.');
   var key = this.getKeyFromMethodNameAndTarget(fn, isStatic);
-  disableMethods[key] = true;
+  this.disableMethodByName(key);
+};
+
+/**
+ * Find a sharedMethod with the given static or prototype method name.
+ *
+ * @param {String} methodName The method name
+ * Find a static or prototype method with the given name.
+ * @returns {SharedMethod}
+ */
+
+SharedClass.prototype.findMethodByName = function(methodName) {
+  return this.methods().filter(function(method) {
+    return method.isDelegateForName(methodName);
+  })[0];
+};
+
+/**
+ * Disable a sharedMethod with the given static or prototype method name.
+ *
+ * @param {String} methodName The method name
+ */
+
+SharedClass.prototype.disableMethodByName = function(methodName) {
+  var disableMethods = this._disabledMethods;
+  disableMethods[methodName] = true;
 };
 
 /**
--- a/lib/shared-method.js
+++ b/lib/shared-method.js
@@ -137,6 +137,12 @@ function SharedMethod(fn, name, sc, options) {
     this.errors = [this.errors];
   }
 
+  if (/^prototype\./.test(name)) {
+    var msg = 'Incorrect API usage. Shared methods on prototypes should be ' +
+      'created via `new SharedMethod(fn, "name", { isStatic: false })`';
+    throw new Error(msg);
+  }
+
   this.stringName = (sc ? sc.name : '') + (isStatic ? '.' : '.prototype.') + name;
 }
 
@@ -605,6 +611,27 @@ SharedMethod.prototype.isDelegateFor = function(suspect, isStatic) {
   }
 
   return false;
+};
+
+/**
+ * Determine if this shared method invokes the given "suspect" function.
+ *
+ * @example
+ * ```js
+ * sharedMethod.isDelegateForName('myMethod'); // check for a static method by name
+ * sharedMethod.isDelegateForName('prototype.myInstMethod'); // instance method by name
+ * ```
+ *
+ * @returns Boolean True if the shared method invokes the given function; false otherwise.
+ */
+
+SharedMethod.prototype.isDelegateForName = function(suspect) {
+  assert(typeof suspect === 'string', 'argument of isDelegateForName should be string');
+
+  var m = suspect.match(/^prototype\.(.*)$/);
+  var isStatic = !m;
+  var baseName = isStatic ? suspect : m[1];
+  return this.isDelegateFor(baseName, isStatic);
 };
 
 /**
--- a/test/shared-class.test.js
+++ b/test/shared-class.test.js
@@ -50,6 +50,7 @@ describe('SharedClass', function() {
       expect(fns).to.contain(SomeClass.staticMethod);
       expect(fns).to.contain(SomeClass.prototype.instanceMethod);
     });
+
     it('returns all methods when includeDisabled is true', function() {
       var sc = new SharedClass('MySharedClass', MySharedClass);
       function MySharedClass() {
@@ -58,8 +59,8 @@ describe('SharedClass', function() {
 
       var inputNames = ['foo', 'bar'];
 
-      sc.defineMethod(inputNames[0], {shared: false});
-      sc.defineMethod(inputNames[1], {shared: true});
+      sc.defineMethod(inputNames[0], {shared: false, isStatic: true});
+      sc.defineMethod(inputNames[1], {shared: true, isStatic: true});
 
       var outputNames = sc.methods({includeDisabled: true}).map(function(m) {
         return m.name;
@@ -67,6 +68,7 @@ describe('SharedClass', function() {
 
       expect(outputNames).to.eql(inputNames);
     });
+
     it('only discovers a function once with aliases', function() {
       function MyClass() {}
       var sc = new SharedClass('some', MyClass);
@@ -81,6 +83,7 @@ describe('SharedClass', function() {
       expect(fns.length).to.equal(1);
       expect(methods[0].aliases.sort()).to.eql(['a', 'b']);
     });
+
     it('discovers multiple functions correctly', function() {
       function MyClass() {}
       var sc = new SharedClass('some', MyClass);
@@ -100,6 +103,7 @@ describe('SharedClass', function() {
         return fn;
       }
     });
+
     it('should skip properties that are model classes', function() {
       var sc = new SharedClass('some', SomeClass);
       function MockModel1() {}
@@ -123,7 +127,7 @@ describe('SharedClass', function() {
       SomeClass.prototype.myMethod = function() {};
       var METHOD_NAME = 'myMethod';
       sc.defineMethod(METHOD_NAME, {
-        prototype: true
+        prototype: true,
       });
       var methods = sc.methods().map(getName);
       expect(methods).to.contain(METHOD_NAME);
@@ -182,6 +186,7 @@ describe('SharedClass', function() {
   describe('sharedClass.find()', function() {
     var sc;
     var sm;
+
     beforeEach(function() {
       sc = new SharedClass('SomeClass', SomeClass);
       SomeClass.prototype.myMethod = function() {};
@@ -190,11 +195,31 @@ describe('SharedClass', function() {
         prototype: true
       });
     });
+
     it('finds sharedMethod for the given function', function() {
       assert(sc.find(SomeClass.prototype.myMethod) === sm);
     });
+
     it('find sharedMethod by name', function() {
       assert(sc.find('myMethod') === sm);
+    });
+  });
+
+  describe('sharedClass.findMethodByName()', function() {
+    it('finds sharedMethod by prototype method name', function() {
+      var sc = new SharedClass('SomeClass', SomeClass);
+      var sm = sc.defineMethod('testMethod', {
+        isStatic: false
+      });
+      assert(sc.findMethodByName('prototype.testMethod') === sm);
+    });
+
+    it('find sharedMethod by static method name', function() {
+      var sc = new SharedClass('SomeClass', SomeClass);
+      var sm = sc.defineMethod('myMethod', {
+        isStatic: true
+      });
+      assert(sc.findMethodByName('myMethod') === sm);
     });
   });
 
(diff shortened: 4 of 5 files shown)


0candy:
@bajtos Please review.

3.0-RELEASE-NOTES.md
```diff
@@ -5,3 +5,14 @@ always describe the impact on users and instructions for upgrading
 applications from 2.x to 3.0.
 
 See also https://github.com/strongloop/loopback/blob/master/3.0-DEVELOPING.md
+
+
+## API call for remote method has been changed.
```
bajtos: I find this little bit confusing, I would expect this is changing the way how remote methods are invoked. Instead, we are changing the API for searching/looking up a remote method by name. Could you please rephrase?

3.0-RELEASE-NOTES.md
```diff
@@ -5,3 +5,14 @@ always describe the impact on users and instructions for upgrading
...
+
+Originally, we have SharedClass.prototype.find method for looking up remote
```
bajtos: Please wrape `SharedClass.prototype.find` in backticks, so that it is rendered as a code. Same for the other instance below.

3.0-RELEASE-NOTES.md
```diff
@@ -5,3 +5,14 @@ always describe the impact on users and instructions for upgrading
...
+methods in 2.x. This is now deprecated in 3.0 and is replaced by 
+SharedClass.prototype.findMethod where you can pass in a static method or a
+prototype method name.
```
bajtos: IMO the description is not detailed enough for readers to understand how to use the new API. Please add an examples of how to use it - one example for a static method and another for a prototype method.

test/shared-method.test.js
```diff
@@ -53,25 +53,22 @@ describe('SharedMethod', function() {
     it('checks by name if a function is going to be invoked', function() {
       var mockSharedClass = { prototype: { myName: myFunction } };
       var sharedMethod = new SharedMethod(myFunction, 'myName', mockSharedClass);
-      assert.equal(sharedMethod.isDelegateFor('myName', false), true);
-      assert.equal(sharedMethod.isDelegateFor('myName', true), false);
```
bajtos: Please preserve the `false` test case:

```
assert.equal(sharedMethod.isDelegateFor('prototype.myName'), false);
```

lib/shared-class.js
```diff
@@ -137,7 +137,7 @@ SharedClass.prototype.methods = function(options) {
 SharedClass.prototype.isMethodEnabled = function(sharedMethod) {
   if (!sharedMethod.shared) return false;
 
-  var key = this.getKeyFromMethodNameAndTarget(sharedMethod.name, sharedMethod.isStatic);
+  var key = sharedMethod.name;
```
bajtos: This is not correct. The constructor of `SharedMethod` is still expecting name + `options.isStatic`. I am surprised this was not caught by unit-tests. Could you please check that we have unit-tests for `isMethodEnabled` for both static and prototype cases, and add them if they are missing?

lib/shared-class.js
```diff
@@ -191,47 +191,34 @@ SharedClass.prototype.resolve = function(resolver) {
 };
 
 /**
- * Find a sharedMethod with the given name or function object.
+ * Find a sharedMethod with the given name or prototype name.
  *
- * @param {String|Function} fn The function or method name
- * @param {Boolean} [isStatic] Required if `fn` is a `String`.
- * Only find a static method with the given name.
+ * @param {String} methodName The method name or prototype method
```
bajtos: Interesting. I didn't realise `find` was supporting both name-based lookup and function-based lookup. I think we should preserve function-based lookup implementation somewhere, as we need it `SharedClass.prototype.methods`. I suppose it's ok to keep it in an private method only.

To simplify upgrade path, I would like to keep `SharedClass.prototype.find` with a deprecation warning (we use `depd` for that). Also to make the distinction between function-based and name-based lookup clear, I am proposing to rename `findMethod` to `findMethodByName`.

```js
var deprecated = require('depd')('strong-remoting'); // at the top

SharedClass.prototype.find = function(fn, isStatic) {
  deprecated('SharedClass.prototype.find is deprecated. Use SharedClass.prototype.findMethodByName instead.');
  // ...
};
```

bajtos:
@0candy Your proposed patch is changing the way how shared methods are internally looked up. I am worried this may have unintended side effect and cause subtle regressions, because the test-coverage of strong-remoting is not sufficient.

When I was writing down the proposal in https://github.com/strongloop/loopback/issues/597#issuecomment-173191509, I didn't realise `find`, `disableMethod` and others were supporting both name-based lookup and function-based lookup. I don't know how many places are relying on the function-based lookup, I am not confident we can safely remove it.

In the light of the above, I am proposing to take a slightly different direction:

Keep `SharedClass.prototype.find` and `SharedClass.prototype.disableMethod`. Add two new methods `SharedClass.prototype.findMethodByName` and `SharedClass.prototype.disableMethodByName`. These new methods should accept a string in the form of "name" for a static method and "prototype.name" for a prototype method. Internally they can delegate the implementation to existing "name+isStatic" methods. Since this is fully backwards-compatible, it should be back-ported to 2.x too (the strong-remoting part only).

The next step is to mark `find` and `disableMethod` as deprecated in 3.0. Then we need to run unit-tests of strong-remoting, loopback and loopback-connector-remote (at least), look for deprecation messages pointing to places using the old API and fix them to use the new API.

bajtos:
Let me explain the reasoning behind my comment above. Breaking changes makes it difficult for existing projects to upgrade to the next major version of LoopBack, thus people tend to stick to whatever legacy LoopBack version they started the project with. On the other hand, we as LoopBack maintainers would like to get as many people running the latest version as possible, so that we don't have to maintain too many old major versions. Therefore we should be very careful whenever a breaking change is introduced, and make sure we make it as easy to upgrade from older versions as possible.

0candy:
@bajtos I agree with you regarding the new direction. To keep the existing methods and add a deprecation note and create two new methods. 
However, after looking into the code, I found that if we internally delegate the implementation of the two new methods to the old ones, there will be issues. As there are lines of code in the internal implementation of the old methods such as getKeyFromMethodNameAndTarget which add 'prototype.' again to the method name that is not static and SharedMethod.prototype.isDelegateFor which looks for isStatic to match the member variable isStatic set in the SharedMethod constructor.

lib/shared-class.js
```diff
@@ -212,12 +215,37 @@ SharedClass.prototype.find = function(fn, isStatic) {
...
 /**
+ * Find a sharedMethod with the given static or prototype method name.
+ *
+ * @param {String} methodName The method name
+ * Find a static or prototype method with the given name.
+ * @returns {SharedMethod}
+ */
+
+SharedClass.prototype.findMethodByName = function(methodName) {
+  var isStatic = methodName.indexOf('prototype.') < 0;
```
bajtos: Please use a stricter check: 

```js
var isStatic = !/^prototype\./.test(methodName);
```

lib/shared-class.js
```diff
@@ -212,12 +215,37 @@ SharedClass.prototype.find = function(fn, isStatic) {
...
+};
+
+/**
+ * Disable a sharedMethod with the given static or prototype method name.
+ *
+ * @param {String} methodName The method name
+ */
+
+SharedClass.prototype.disableMethodByName = function(methodName) {
+  this.disableMethod(methodName);
```
bajtos: If this rather hacky. IIUC, it's relying on the fact that `getKeyFromMethodNameAndTarget` will pass through `prototype.name`.

test/shared-class.test.js
```diff
@@ -198,6 +198,25 @@ describe('SharedClass', function() {
     });
   });
 
+  describe('sharedClass.findMethodByName()', function() {
+    var sc;
+    var sm;
```
bajtos: What's the point of declaring the variables here, when they are always reinitialized in each test? Please keep them test-local.

test/shared-class.test.js
```diff
@@ -198,6 +198,25 @@ describe('SharedClass', function() {
...
+    it('finds sharedMethod by prototype name', function() {
+      sc = new SharedClass('SomeClass', SomeClass);
+      var PROTOTYPE_METHOD_NAME = 'SomeClass.prototype.testMethod';
+      sm = sc.defineMethod(PROTOTYPE_METHOD_NAME);
+      assert(sc.findMethodByName('SomeClass.prototype.testMethod') === sm);
+    });
+    it('find sharedMethod by name', function() {
```
bajtos: Please add an empty line to visually delimit test cases.

bajtos: Also modify the test name to communicate that this test is checking static method.

test/shared-class.test.js
```diff
@@ -198,6 +198,25 @@ describe('SharedClass', function() {
...
   });
 
+  describe('sharedClass.findMethodByName()', function() {
+    var sc;
+    var sm;
+    it('finds sharedMethod by prototype name', function() {
+      sc = new SharedClass('SomeClass', SomeClass);
+      var PROTOTYPE_METHOD_NAME = 'SomeClass.prototype.testMethod';
+      sm = sc.defineMethod(PROTOTYPE_METHOD_NAME);
+      assert(sc.findMethodByName('SomeClass.prototype.testMethod') === sm);
```
bajtos: Either use PROTOTYPE_METHOD_NAME here too, or else get rid (inline all usages) of PROTOTYPE_METHOD_NAME.

I would personally not bother with defining a constant for this string.

bajtos:
@0candy 

> However, after looking into the code, I found that if we internally delegate the implementation of the two new methods to the old ones, there will be issues. As there are lines of code in the internal implementation of the old methods such as getKeyFromMethodNameAndTarget which add 'prototype.' again to the method name that is not static and SharedMethod.prototype.isDelegateFor which looks for isStatic to match the member variable isStatic set in the SharedMethod constructor.

My main issue with the current solution is that when the user calls a new method, it will still print a deprecation warning, because new methods are delegating to old, deprecated methods.

We need to reorganize the code so that both old and new methods call shared code that is not deprecated. Some of the possible approaches: 1) move the implementation to new methods, rework old deprecated methods to call the new methods 2) move the core of the implementation into shared private methods, call these private methods from both old and new public methods. I think 1) will be better in this case, but I don't mind either - please figure this out yourself.

> For this test case, I had to set isStatic for it to pass.
https://github.com/strongloop/strong-remoting/pull/275/files#diff-97fe2c831fe95a92fc8828ad0431d6baR210

Yes, that's expected. The code for creating (defining) new shared methods does not support "prototype.name" notation. I am not sure if this is something to change too. Can we discuss it outside of this pull request please?

Could you please modify all test to explicitly pass `isStatic:true` or `isStatic:false` to make it clear what kind of methods they are creating?

lib/shared-class.js
```diff
@@ -212,9 +214,58 @@ SharedClass.prototype.find = function(fn, isStatic) {
...
+
+SharedClass.prototype._findSharedMethod = function(fn, isStatic) {
+  var methods = this.methods();
+  return find(methods, fn, isStatic);
+};
+
+/**
+ * Internal implementation to disable a sharedMethod.
+ *
+ * @param {String} methodOrFnName The function or method name
```
bajtos: I would expect that in this private function, the argument must be always a string (method name):
1) `this._disabledMethods` is indexed by strings IIRC 
2) the deprecated public method `disableMethods` should take care of conversion from function to method name.

lib/shared-class.js
```diff
@@ -225,7 +276,11 @@ SharedClass.prototype.disableMethod = function(fn, isStatic) {
  */
 
 SharedClass.prototype.getKeyFromMethodNameAndTarget = function(name, isStatic) {
-  return (isStatic ? '' : 'prototype.') + name;
+  var addPrototypeString = !/^prototype\./.test(name);
```
bajtos: Interesting. So IIUC, this is adding support for `getKeyFromMethodNameAndTarget('prototype.name')`. I am wondering where is it used, why do we need to make this change?

lib/shared-method.js
```diff
@@ -137,7 +137,9 @@ function SharedMethod(fn, name, sc, options) {
     this.errors = [this.errors];
   }
 
-  this.stringName = (sc ? sc.name : '') + (isStatic ? '.' : '.prototype.') + name;
+  var addPrototypeString = !/^prototype\./.test(name);
+  this.stringName = (sc ? sc.name : '') +
+    (!isStatic && addPrototypeString ? '.prototype.' : '.') + name;
```
bajtos: ditto. Why would you like to support `new SharedMethod(fn, 'prototype.name', ...)`?

bajtos: Is this based on my earlier [comment](https://github.com/strongloop/strong-remoting/pull/275#issuecomment-176144736)?

> The code for creating (defining) new shared methods does not support "prototype.name" notation. I am not sure if this is something to change too. Can we discuss it outside of this pull request please?

bajtos:
@0candy Nice, the patch looks much better now. Please rebase it on top of the current master and fix merge conflicts (if any).

I left few comments to address, see above. Perhaps we can discuss the change in `getKeyFromMethodNameAndTarget` and `SharedMethod` online?

Last but not least: the CI build is failing on few machines, please check and fix.

test/shared-class.test.js
```diff
@@ -243,6 +268,26 @@ describe('SharedClass', function() {
...
+      var sm = sc.defineMethod(METHOD_NAME, {isStatic: true});
+      sc.disableMethodByName(METHOD_NAME);
+      var methods = sc.methods().map(getName);
+      expect(methods).to.not.contain(METHOD_NAME);
+    });
+
+    it('excludes disabled prototype methods from the method list', function() {
+      var INST_METHOD_NAME = 'prototype.instTestMethod';
+      var sc = new SharedClass('SomeClass', SomeClass);
+      var sm = sc.defineMethod(INST_METHOD_NAME, {isStatic: false});
```
bajtos: This is incorrect API usage which creates the situation when you need to add checks whether method name contains `prototype.` prefix.

This is the correct way for registering a new prototype method:

```js
sc.defineMethod('instTestMethod', { isStatic: false });
```

lib/shared-method.js
bajtos: As far as this patch is concerned, I think we should not add support for `name:'prototype.method'`, as I already mentioned before, so this block should be reverted.

However, in order to prevent other users from running into the situation described in https://github.com/strongloop/strong-remoting/pull/275/files#r51434265, I am proposing to add a check here instead and throw an error when the method name starts with `prototype.`.

```js
if (/^prototype\./.test(name)) {
  var msg = 'Incorrect API usage. Shared methods on prototypes should be ' +
    'created via "new SharedMethod(fn, "name", { isStatic: false })';
  throw new Error(msg);
}

File: 3.0-RELEASE-NOTES.md [hunk 16]
```diff
@@ -5,3 +5,24 @@ always describe the impact on users and instructions for upgrading
 applications from 2.x to 3.0.
 
 See also https://github.com/strongloop/loopback/blob/master/3.0-DEVELOPING.md
+
+
+## Invokation for remote method has been changed.
```
bajtos: Nitpick: I think the correct spelling is "invocation" (with "c", not "k").

lib/shared-method.js
```diff
@@ -137,6 +137,12 @@ function SharedMethod(fn, name, sc, options) {
     this.errors = [this.errors];
   }
 
+  if (/^prototype\./.test(name)) {
+    var msg = 'Incorrect API usage. Shared methods on prototypes should be ' +
+      'created via "new SharedMethod(fn, "name", { isStatic: false })';
```
bajtos: I think the quote characters are a bit off here - you use the same quote char for delimiting the code snippet and strings inside the snippet, and also the closing `"` is missing. Let's do this instead:

```
'created via `new SharedMethod(fn, "name", { isStatic: false })`';
```

lib/shared-method.js
```diff
@@ -608,6 +614,29 @@ SharedMethod.prototype.isDelegateFor = function(suspect, isStatic) {
...
+ *
+ * @returns Boolean True if the shared method invokes the given function; false otherwise.
+ */
+
+SharedMethod.prototype.isDelegateForName = function(suspect) {
+  var isStatic = !/^prototype\./.test(suspect);
+  var baseName = suspect;
+  if (!isStatic) {
+    baseName = suspect.substring(10);
+  }
```
bajtos: This is sort of repeating, as `10` is tied to the value in the regexp on the first line.

I think this is more elegant:

```js
var m = suspect.match(/^prototype\.(.*)$/);
var isStatic = !m;
var baseName = isStatic ? suspect : m[1];
```

test/shared-class.test.js
```diff
@@ -243,6 +268,26 @@ describe('SharedClass', function() {
...
+      expect(methods).to.not.contain(METHOD_NAME);
+    });
+
+    it('excludes disabled prototype methods from the method list', function() {
+      var INST_METHOD_NAME = 'prototype.instTestMethod';
+      var sc = new SharedClass('SomeClass', SomeClass);
+      var sm = sc.defineMethod('instTestMethod', {isStatic: false});
+      sc.disableMethodByName(INST_METHOD_NAME);
+      var methods = sc.methods().map(getName);
+      expect(methods).to.not.contain(INST_METHOD_NAME);
```
bajtos: Are we sure `getName` returns the name with the `prototype` suffix? I think the opposite is the case.

To be sure, could you please add another check

```js
expect(methods).to.not.contain('instTestMethod');
```

lib/shared-method.js
```diff
@@ -608,6 +614,27 @@ SharedMethod.prototype.isDelegateFor = function(suspect, isStatic) {
 };
 
 /**
+ * Determine if this shared method invokes the given "suspect" function.
+ *
+ * @example
+ * ```js
+ * sharedMethod.isDelegateForName(myClass.myMethod); // pass a function
+ * sharedMethod.isDelegateForName(myClass.prototype.myInstMethod);
```
bajtos: I think `suspect` arg should be always a string, shouldn't it?

Could you please add an `assert` as the first thing in the method to verify this precondition?

bajtos:
@0candy one more comment - see above, also please add a unit-test for the new method `isDelegateForName`.

bajtos:
@0candy the comment above is minor and can be ignored, but please do check why the CI build is failing.

bajtos:
Other than that, the patch LGTM. You can land it with no further review once the CI build is green(ish).

PR #277, opened 2016-01-28
Implement vertical slice: Custom router

gunjpan:
This PR is a POC.

Connect to https://github.com/strongloop-internal/scrum-loopback/issues/714

### DIFF
--- a/lib/rest-adapter.js
+++ b/lib/rest-adapter.js
@@ -116,7 +116,7 @@ RestAdapter.prototype.invoke = function(method, ctorArgs, args, callback) {
   var restMethod = this.getRestMethodByName(method);
   var invocation = new HttpInvocation(
     restMethod, ctorArgs, args, this.connection, remotes.auth
-  );
+    );
   var ctx = { req: invocation.createRequest() };
   var scope = remotes.getScope(ctx, restMethod);
   remotes.execHooks('before', restMethod, scope, ctx, function(err) {
@@ -229,13 +229,13 @@ RestAdapter.prototype.createHandler = function() {
 
   // Set strict to be `false` so that anything `JSON.parse()` accepts will be parsed
   debug('remoting options: %j', this.remotes.options);
-  var urlencodedOptions = this.remotes.options.urlencoded || {extended: true};
+  var urlencodedOptions = this.remotes.options.urlencoded || { extended: true };
   if (urlencodedOptions.extended === undefined) {
     urlencodedOptions.extended = true;
   }
-  var jsonOptions = this.remotes.options.json || {strict: false};
+  var jsonOptions = this.remotes.options.json || { strict: false };
   var corsOptions = this.remotes.options.cors;
-  if (corsOptions === undefined) corsOptions = {origin: true, credentials: true};
+  if (corsOptions === undefined) corsOptions = { origin: true, credentials: true };
 
   // Optimize the cors handler
   var corsHandler = function(req, res, next) {
@@ -260,24 +260,21 @@ RestAdapter.prototype.createHandler = function() {
   // mount rest route handler
   var customRouter = new RestRouter();
   root.use(customRouter.handle());
-  
+
   classes.forEach(function(restClass) {
     var router = express.Router();
     var className = restClass.sharedClass.name;
 
     debug('registering REST handler for class %j', className);
-    
+
     // Mount the remoteClass router on all class routes.
     restClass
       .routes
-      .forEach(function (route) {
+      .forEach(function(route) {
         debug('    at %s', route.path);
         root.use(route.path, router);
-        route.fullPath = restClass.getPath(); //add fullPath for Trie insertion
-        customRouter.registerPathAndHandlers(route);
-        //Do we need to register above class route?if yes, what's all method verb?
       });
-      
+
     var methods = [];
     // Register handlers for all shared methods of this class sharedClass
     restClass
@@ -286,9 +283,9 @@ RestAdapter.prototype.createHandler = function() {
         var sharedMethod = restMethod.sharedMethod;
         debug('    method %s', sharedMethod.stringName);
         restMethod.routes.forEach(function(route) {
-          //add fullPath for Trie insertion          
+          //add fullPath for Trie insertion
           route.fullPath = joinPaths(restClass.getPath(), route.path);
-          methods.push({route: route, method: sharedMethod});
+          methods.push({ route: route, method: sharedMethod });
         });
       });
 
@@ -304,7 +301,7 @@ RestAdapter.prototype.createHandler = function() {
       // Do not allow other middleware to invade our URL space.
       router.use(RestAdapter.remoteMethodNotFoundHandler(className));
     }
-    
+
   });
 
   if (handleUnknownPaths) {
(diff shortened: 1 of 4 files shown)


lib/rest-adapter.js
```diff
@@ -256,12 +257,27 @@ RestAdapter.prototype.createHandler = function() {
...
     debug('registering REST handler for class %j', className);
-
+    
+    // Mount the remoteClass router on all class routes.
+    restClass
+      .routes
+      .forEach(function (route) {
+        debug('    at %s', route.path);
+        root.use(route.path, router);
+        route.fullPath = restClass.getPath(); //add fullPath for Trie insertion
```
bajtos: This is smelly. Surely the full route of the path is not just the path of the class? Perhaps `route.classPath` would be a better property name?

lib/rest-adapter.js
```diff
@@ -256,12 +257,27 @@ RestAdapter.prototype.createHandler = function() {
...
+        customRouter.registerPathAndHandlers(route);
+        //Do we need to register above class route?if yes, what's all method verb?
```
bajtos: > Do we need to register above class route?if yes, what's all method verb?

No, this does not make sense to me. There is no such thing as a class route. Only methods can contribute routes.

lib/rest-adapter.js
```diff
@@ -286,15 +304,7 @@ RestAdapter.prototype.createHandler = function() {
...
-
-    // Mount the remoteClass router on all class routes.
-    restClass
-      .routes
-      .forEach(function(route) {
-        debug('    at %s', route.path);
-        root.use(route.path, router);
-      });
-
+
```
bajtos: Extra empty line, please remove.

lib/rest-router.js
```diff
@@ -0,0 +1,36 @@
+module.exports = RestRouter;
+
+var Trie = require('./trie');
+
+function RestRouter(opts) {
+  this.Trie = new Trie();
```
bajtos: Variable and property names should start with a lower-case letter, i.e. `this.trie`.

lib/rest-router.js
```diff
@@ -0,0 +1,36 @@
...
+  this.opts = opts;
```
bajtos: Don't abbreviate here, `this.options = options`.

lib/rest-router.js
```diff
@@ -0,0 +1,36 @@
...
+    var methods = Trie.find(path);
+    if (methods && verb in methods) {
+      methods[verb](req, res, next);
+    }
+    else next();
+  }
+}
+
+function getPath(path) {
+  if (!this.opts || !(this.opts.caseSensitive)) {
```
bajtos: `this` is undefined in this function. I am proposing this:

```js
function normalizePath(path, options) {
  // your implementation, no change needed
};

// usage on L22:
var path = normalizePath(path, this.options);
```

lib/trie.js
```diff
@@ -0,0 +1,57 @@
...
+      if (i + 1 === parts.length) {
+        node[parts[i]].methods = node[parts[i]].methods || {};
+        node[parts[i]].methods[verb] = handler;
+      }
+      node = node[parts[i]];
+    }
+  }
+  else { // if we would ever encounter this situation? 
+    node.methods = node.methods || {};
+    node.methods[verb] = handler;
```
bajtos: What happens when there is more than one handler registered for the same verb+path? Please check the current router does and implement the same behaviour. Don't forget to add unit-test.

lib/trie.js
```diff
@@ -0,0 +1,57 @@
...
+
+//insert new nodes to Trie
+Trie.prototype.add = function (path, verb, handler) {
+  var node = this;
+  var parts = getPathParts(path);
+
+  if (parts.length) {
+    for (var i = 0; i < parts.length; i++) {
+      if (!node || !(parts[i] in node)) {
+        node[parts[i]] = new Trie();
```
bajtos: The first part of the condition will not work, as it distills to `if (!node) node[pars[i]] = new Trie()` - i.e. attempts to assign to an undefined variable.

lib/trie.js
```diff
@@ -0,0 +1,57 @@
...
+      }
+      if (i + 1 === parts.length) {
+        node[parts[i]].methods = node[parts[i]].methods || {};
+        node[parts[i]].methods[verb] = handler;
```
bajtos: L18-19 and L25-26 are duplicated, please extract a shared private method.

An example of a possible API in action:

```js
node[parts[i]].addVerbHandler(verb, handler);
```

lib/trie.js
```diff
@@ -0,0 +1,57 @@
...
+
+//Look for matching path and return handler methods for a match
+Trie.prototype.find = function (path) {
+  var node = this;
+  var parts = getPathParts(path);
+  for (var i = 0; i < parts.length; i++) {
+    if (!(parts[i] in node)) {
+      return false;
+    }
+    else {
```
bajtos: Coding style: merge L38 and L39 into a single line.

```js
} else {
```

lib/trie.js
```diff
@@ -0,0 +1,57 @@
...
+      if (i + 1 === parts.length)
+        return node[parts[i]].methods;
+    }
+    node = node[parts[i]];
+  }
+}
+
+function getPathParts(path) {
+  path = path.trim();
```
bajtos: Is express router trimming whitespace too?

bajtos:
The direction looks very reasonable to me. I am not going to review implementation details in respect to edge cases now, I need to see a test suite first.

lib/rest-adapter.js
gunjpan: @bajtos: so, removing L277 would take care of the comments above

lib/trie.js
gunjpan: @bajtos: I found that for a single `verb+path`, there is only one entry point in the express router. The router creates a stack of handler functions, which only invokes the handler, which was registered first by `router[method](path, handler)`, for the path. Other handlers won't see the light if they're not invoked if `request/response` cycle ends before `next()`.
So, for example, in the code below:
```
router.get('/planets', handlerOne());
router.get('/planets', handlerTwo());
```
only `handlerOne()` will be executed if it sends the `response` and doesn't reach the `next()`.
Thus, I believe, we will need to accommodate this case.
Thus, in the code above, at L19, an Array would be helpful to index registering of handlers for same verb+path:
`  node[parts[i]].methods[verb] = handler.push();`

Your thoughts please!

bajtos: Ah, of course, you are right. Your proposal to keep all handlers in an array makes perfect sense.

bajtos: Please make sure to add a unit-test for this case.

lib/rest-adapter.js
bajtos: Makes sense.

lib/trie.js
gunjpan: @bajtos: express router uses [`parseurl` module](https://github.com/pillarjs/parseurl/blob/master/index.js#L23) which uses a regular expression for verifying the url pattern (this pattern discards match for whitespace).

Using `parseurl` will be the right thing to do.

gunjpan:
Closing this as this POC is used in Iteration-1 of custom rest-router implementation #282 
No further review is required for this PR (keeping the branch for now)

PR #278, opened 2016-01-28
Extract ContextBase and ctx.getScope()

bajtos:
Create `ContextBase` - a base class for all Context objects where we can put methods shared by all contexts.

Move `remotes.getScope` to `ContextBase.prototype.getScope` as it belongs there more naturally and will make it easier to access scope in hooks that can access the `ctx` object only.

Connect to strongloop-internal/scrum-loopback#729

@raymondfeng @ritch please review

### DIFF
--- a/lib/context-base.js
+++ b/lib/context-base.js
@@ -1,0 +1,26 @@
+'use strict';
+
+var EventEmitter = require('events').EventEmitter;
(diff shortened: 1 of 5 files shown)


ritch:
LGTM... meant to do this quite a while ago. :+1:

PR #282, opened 2016-02-04
Implement a faster REST router

gunjpan:
Faster  REST router implementation - iteration 1
Includes: 
- Handler method registration and invocation for simple REST paths i.e. `/api/[model]` & `/api/[model]/[method]`

Excludes: 
- Handler method registration and invocation for REST paths with `:parameters` e.x. `/api/[model]/:id`
- Handlers with more than three arguments for example, `handlerFn(err, req, res, next){..}`

Connect to https://github.com/strongloop-internal/scrum-loopback/issues/730

### DIFF
--- a/lib/rest-adapter.js
+++ b/lib/rest-adapter.js
@@ -282,7 +282,7 @@ RestAdapter.prototype.createHandler = function() {
         var sharedMethod = restMethod.sharedMethod;
         debug('    method %s', sharedMethod.stringName);
         restMethod.routes.forEach(function(route) {
-          //add fullPath for Trie insertion
+          // add fullPath for Trie insertion
           route.fullPath = joinPaths(restClass.getPath(), route.path);
           methods.push({ route: route, method: sharedMethod });
         });
--- a/lib/rest-adapter.js
+++ b/lib/rest-adapter.js
@@ -434,11 +434,11 @@ RestAdapter.prototype._registerMethodRouteHandlers = function(router,
     verb = 'delete';
   }
 
-  if (fastRouter.canRegister(route, handler)) {
-    handler.methodName = sharedMethod.name;
-    fastRouter[verb](route, handler);
-  }
-  router[verb](route.path, handler);
+  // if (fastRouter.canRegister(route, handler)) {
+  handler.methodName = sharedMethod.name;
+  fastRouter[verb](route, handler);
+  // }
+  // router[verb](route.path, handler);
 };
 
 RestAdapter.prototype._createStaticMethodHandler = function(sharedMethod) {
--- a/lib/rest-adapter.js
+++ b/lib/rest-adapter.js
@@ -28,6 +28,7 @@ var async = require('async');
 var HttpInvocation = require('./http-invocation');
 var ContextBase = require('./context-base');
 var HttpContext = require('./http-context');
+var RestRouter = require('./rest-router');
 
 var json = bodyParser.json;
 var urlencoded = bodyParser.urlencoded;
@@ -263,6 +264,10 @@ RestAdapter.prototype.createHandler = function() {
 
   var handleUnknownPaths = this._shouldHandleUnknownPaths();
 
+  // Mount rest-router
+  var customRouter = new RestRouter();
+  root.use(customRouter.handle());
+
   classes.forEach(function(restClass) {
     var router = express.Router();
     var className = restClass.sharedClass.name;
@@ -277,7 +282,13 @@ RestAdapter.prototype.createHandler = function() {
         var sharedMethod = restMethod.sharedMethod;
         debug('    method %s', sharedMethod.stringName);
         restMethod.routes.forEach(function(route) {
+<<<<<<< HEAD
           methods.push({ route: route, method: sharedMethod });
+=======
+          //add fullPath for Trie insertion
+          route.fullPath = joinPaths(restClass.getPath(), route.path);
+          methods.push({route: route, method: sharedMethod});
+>>>>>>> cd24be5... Implement rest router from POC.
         });
       });
 
@@ -285,7 +296,7 @@ RestAdapter.prototype.createHandler = function() {
     methods.sort(sortRoutes);
 
     methods.forEach(function(m) {
-      adapter._registerMethodRouteHandlers(router, m.method, m.route);
+      adapter._registerMethodRouteHandlers(router, m.method, m.route, customRouter);
     });
 
     if (handleUnknownPaths) {
@@ -414,7 +425,8 @@ RestAdapter.errorHandler = function(options) {
 
 RestAdapter.prototype._registerMethodRouteHandlers = function(router,
                                                               sharedMethod,
-                                                              route) {
+                                                              route,
+                                                              customRouter) {
   var handler = sharedMethod.isStatic ?
     this._createStaticMethodHandler(sharedMethod) :
     this._createPrototypeMethodHandler(sharedMethod);
@@ -424,6 +436,14 @@ RestAdapter.prototype._registerMethodRouteHandlers = function(router,
   if (verb === 'del') {
     // Express 4.x only supports delete
     verb = 'delete';
+  }
+
+  //For now: exclude paths with parameters like :id
+  //and handler functions with err obj i.e. (err, req,res,next)
+  if (!route.path.includes(':') && handler.length < 4) {
+    handler.methodName = sharedMethod.name;
+    handler.path = route.fullPath;
+    customRouter.registerPathAndHandlers({ verb: verb, fullPath: route.fullPath }, handler);
   }
   router[verb](route.path, handler);
 };
--- a/lib/rest-adapter.js
+++ b/lib/rest-adapter.js
@@ -265,8 +265,8 @@ RestAdapter.prototype.createHandler = function() {
   var handleUnknownPaths = this._shouldHandleUnknownPaths();
 
   // Mount rest-router
-  var customRouter = new RestRouter();
-  root.use(customRouter.handle());
+  var fastRouter = new RestRouter();
+  root.use(fastRouter);
 
   classes.forEach(function(restClass) {
     var router = express.Router();
@@ -282,13 +282,9 @@ RestAdapter.prototype.createHandler = function() {
         var sharedMethod = restMethod.sharedMethod;
         debug('    method %s', sharedMethod.stringName);
         restMethod.routes.forEach(function(route) {
-<<<<<<< HEAD
-          methods.push({ route: route, method: sharedMethod });
-=======
           //add fullPath for Trie insertion
           route.fullPath = joinPaths(restClass.getPath(), route.path);
-          methods.push({route: route, method: sharedMethod});
->>>>>>> cd24be5... Implement rest router from POC.
+          methods.push({ route: route, method: sharedMethod });
         });
       });
 
@@ -296,7 +292,7 @@ RestAdapter.prototype.createHandler = function() {
     methods.sort(sortRoutes);
 
     methods.forEach(function(m) {
-      adapter._registerMethodRouteHandlers(router, m.method, m.route, customRouter);
+      adapter._registerMethodRouteHandlers(router, m.method, m.route, fastRouter);
     });
 
     if (handleUnknownPaths) {
@@ -426,7 +422,7 @@ RestAdapter.errorHandler = function(options) {
 RestAdapter.prototype._registerMethodRouteHandlers = function(router,
                                                               sharedMethod,
                                                               route,
-                                                              customRouter) {
+                                                              fastRouter) {
   var handler = sharedMethod.isStatic ?
     this._createStaticMethodHandler(sharedMethod) :
     this._createPrototypeMethodHandler(sharedMethod);
@@ -438,12 +434,9 @@ RestAdapter.prototype._registerMethodRouteHandlers = function(router,
     verb = 'delete';
   }
 
-  //For now: exclude paths with parameters like :id
-  //and handler functions with err obj i.e. (err, req,res,next)
-  if (!route.path.includes(':') && handler.length < 4) {
+  if (fastRouter.canRegister(route, handler)) {
     handler.methodName = sharedMethod.name;
-    handler.path = route.fullPath;
-    customRouter.registerPathAndHandlers({ verb: verb, fullPath: route.fullPath }, handler);
+    fastRouter[verb](route, handler);
   }
   router[verb](route.path, handler);
 };
--- a/lib/rest-adapter.js
+++ b/lib/rest-adapter.js
@@ -438,7 +438,7 @@ RestAdapter.prototype._registerMethodRouteHandlers = function(router,
   handler.methodName = sharedMethod.name;
   fastRouter[verb](route, handler);
   // }
-  // router[verb](route.path, handler);
+  router[verb](route.path, handler);
 };
 
 RestAdapter.prototype._createStaticMethodHandler = function(sharedMethod) {
--- a/lib/rest-router.js
+++ b/lib/rest-router.js
@@ -3,7 +3,7 @@ var url = require('url');
 var http = require('http');
 var Trie = require('./trie');
 
-var HTTP_METHODS = getCurrentNodeMethods();
+var HTTP_METHODS = require('methods');
 
 var RestRouter = module.exports = function(options) {
   var opts = options || {};
@@ -56,11 +56,4 @@ function normalizePath(path, options) {
     path = path.toLowerCase();
   }
   return path;
-}
-
-// get list of all HTTP methods
-function getCurrentNodeMethods() {
-  return http.METHODS && http.METHODS.map(function lowerCaseMethod(method) {
-    return method.toLowerCase();
-  });
 }
--- a/lib/rest-router.js
+++ b/lib/rest-router.js
@@ -4,6 +4,7 @@ var http = require('http');
 var Trie = require('./trie');
 
 var methods = getCurrentNodeMethods();
+
 var RestRouter = module.exports = function(options) {
   var opts = options || {};
   var router = function(req, res, next) {
@@ -45,9 +46,9 @@ RestRouter.handle = function(req, res, next) {
 // For now: exclude paths with parameters like :id
 // and handler functions with err obj i.e. (err, req, res, next)
 
-RestRouter.canRegister = function(route, handler) {
-  return !(route.path.includes(':') && handler.length < 4);
-};
+// RestRouter.canRegister = function(route, handler) {
+//   return !(route.path.includes(':') && handler.length < 4);
+// };
 
 function normalizePath(path, options) {
   path = url.parse(path).pathname;
--- a/lib/rest-router.js
+++ b/lib/rest-router.js
@@ -5,7 +5,6 @@ var Trie = require('./trie');
 
 var methods = getCurrentNodeMethods();
 var RestRouter = module.exports = function(options) {
-
   var opts = options || {};
   var router = function(req, res, next) {
     router.handle(req, res, next);
--- a/lib/rest-router.js
+++ b/lib/rest-router.js
@@ -19,7 +19,7 @@ var RestRouter = module.exports = function(options) {
 };
 
 // express style API for registering route handlers: ex. route[method](path, handler)
-// add verb:hanlder for path to trie DS
+// add verb:handler for path to trie DS
 HTTP_METHODS.concat('all').forEach(function(method) {
   RestRouter[method] = function(route, handler) {
     route.fullPath = normalizePath(route.fullPath, this.opts);
--- a/lib/rest-router.js
+++ b/lib/rest-router.js
@@ -3,7 +3,7 @@ var url = require('url');
 var http = require('http');
 var Trie = require('./trie');
 
-var methods = getCurrentNodeMethods();
+var HTTP_METHODS = getCurrentNodeMethods();
 
 var RestRouter = module.exports = function(options) {
   var opts = options || {};
@@ -20,7 +20,7 @@ var RestRouter = module.exports = function(options) {
 
 // express style API for registering route handlers: ex. route[method](path, handler)
 // add verb:hanlder for path to trie DS
-methods.concat('all').forEach(function(method) {
+HTTP_METHODS.concat('all').forEach(function(method) {
   RestRouter[method] = function(route, handler) {
     route.fullPath = normalizePath(route.fullPath, this.opts);
     var trie = this._trie;
--- a/lib/rest-router.js
+++ b/lib/rest-router.js
@@ -1,31 +1,53 @@
-module.exports = RestRouter;
 
 var url = require('url');
+var http = require('http');
 var Trie = require('./trie');
 
-function RestRouter(options) {
-  this.trie = new Trie();
-  this.options = options;
-}
+var methods = getCurrentNodeMethods();
+var RestRouter = module.exports = function(options) {
 
-//register path and related handler to Trie ds
-RestRouter.prototype.registerPathAndHandlers = function(route, handler) {
-  var path = normalizePath(route.fullPath);
-  var trie = this.trie;
-  trie = trie.add(path, route.verb, handler);
+  var opts = options || {};
+  var router = function(req, res, next) {
+    router.handle(req, res, next);
+  };
+
+  router.__proto__ = RestRouter;
+  router._trie = new Trie();
+  router.options = opts;
+
+  return router;
 };
 
-//handle request, match path and if found, invoke handler method
-RestRouter.prototype.handle = function() {
-  var trie = this.trie;
-  return function(req, res, next) {
-    var verb = req.method.toLowerCase() || 'all';
-    var path = normalizePath(req.url, this.options);
-    var methods = trie.find(path);
-    if (methods && verb in methods) {
-      methods[verb](req, res, next);
-    } else next();
+// express style API for registering route handlers: ex. route[method](path, handler)
+// add verb:hanlder for path to trie DS
+methods.concat('all').forEach(function(method) {
+  RestRouter[method] = function(route, handler) {
+    route.fullPath = normalizePath(route.fullPath, this.opts);
+    var trie = this._trie;
+    trie = trie.add(route, handler);
+    return this;
   };
+});
+
+// handle request, match path and if found, invoke handler method
+RestRouter.handle = function(req, res, next) {
+  var trie = this._trie;
+  var verb = req.method.toLowerCase();
+  var path = normalizePath(req.url, this.options);
+  var methods = trie.find(path);
+
+  if (methods && verb in methods) {
+    methods[verb](req, res, next);
+  } else {
+    next();
+  }
+};
+
+// For now: exclude paths with parameters like :id
+// and handler functions with err obj i.e. (err, req, res, next)
+
+RestRouter.canRegister = function (route, handler) {
+  return !(route.path.includes(':') && handler.length < 4)
 };
 
 function normalizePath(path, options) {
@@ -34,4 +56,11 @@ function normalizePath(path, options) {
     path = path.toLowerCase();
   }
   return path;
+}
+
+// get list of all HTTP methods
+function getCurrentNodeMethods() {
+  return http.METHODS && http.METHODS.map(function lowerCaseMethod(method) {
+    return method.toLowerCase();
+  });
 }
--- a/lib/rest-router.js
+++ b/lib/rest-router.js
@@ -34,7 +34,8 @@ RestRouter.handle = function(req, res, next) {
   var trie = this._trie;
   var verb = req.method.toLowerCase();
   var path = normalizePath(req.url, this.options);
-  var methods = trie.find(path);
+  var methods = trie.find(path, req);
+  console.log('req.params: ', req.params);
 
   if (methods && verb in methods) {
     methods[verb](req, res, next);
--- a/lib/rest-router.js
+++ b/lib/rest-router.js
@@ -46,8 +46,8 @@ RestRouter.handle = function(req, res, next) {
 // For now: exclude paths with parameters like :id
 // and handler functions with err obj i.e. (err, req, res, next)
 
-RestRouter.canRegister = function (route, handler) {
-  return !(route.path.includes(':') && handler.length < 4)
+RestRouter.canRegister = function(route, handler) {
+  return !(route.path.includes(':') && handler.length < 4);
 };
 
 function normalizePath(path, options) {
--- a/lib/rest-router.js
+++ b/lib/rest-router.js
@@ -1,0 +1,37 @@
+module.exports = RestRouter;
+
+var url = require('url');
+var Trie = require('./trie');
+
+function RestRouter(options) {
+  this.trie = new Trie();
+  this.options = options;
+}
+
+//register path and related handler to Trie ds
+RestRouter.prototype.registerPathAndHandlers = function(route, handler) {
+  var path = normalizePath(route.fullPath);
+  var trie = this.trie;
+  trie = trie.add(path, route.verb, handler);
+};
+
+//handle request, match path and if found, invoke handler method
+RestRouter.prototype.handle = function() {
+  var trie = this.trie;
+  return function(req, res, next) {
+    var verb = req.method.toLowerCase() || 'all';
+    var path = normalizePath(req.url, this.options);
+    var methods = trie.find(path);
+    if (methods && verb in methods) {
+      methods[verb](req, res, next);
+    } else next();
+  };
+};
+
+function normalizePath(path, options) {
+  path = url.parse(path).pathname;
+  if (!options || !(options.caseSensitive)) {
+    path = path.toLowerCase();
+  }
+  return path;
+}
--- a/lib/trie.js
+++ b/lib/trie.js
@@ -1,41 +1,50 @@
+
 module.exports = Trie;
 
 function Trie() {
   this.methods = {};
 }
 
-//insert new nodes to Trie
-Trie.prototype.add = function(path, verb, handler) {
+// insert new nodes to Trie
+Trie.prototype.add = function(route, handler) {
   var node = this;
-  var parts = getPathParts(path);
-  if (parts.length) {
-    for (var i = 0; i < parts.length; i++) {
-      if (!(parts[i] in node)) {
-        node[parts[i]] = new Trie();
-      }
-      if (i + 1 === parts.length) {
-        addVerbHandler(node[parts[i]], verb, handler);
-      }
-      node = node[parts[i]];
-    }
-  } else { // if we would ever encounter this situation?
-    addVerbHandler(node, verb, handler);
+  var parts = getPathParts(route.fullPath);
+  var segment;
+
+  if (!parts.length) {
+    addVerbHandler(node, route, handler);
+    return this;
   }
+  for (var i = 0; i < parts.length; i++) {
+    segment = parts[i];
+    if (!(segment in node)) {
+      node[segment] = new Trie();
+    }
+    if (i + 1 === parts.length) {
+      addVerbHandler(node[segment], route, handler);
+    }
+    node = node[segment];
+  }
+
   return this;
 };
 
-//Look for matching path and return handler method for a match
+// Look for matching path and return methods for a match
 Trie.prototype.find = function(path) {
   var node = this;
   var parts = getPathParts(path);
+  var segment;
+
   for (var i = 0; i < parts.length; i++) {
-    if (!(parts[i] in node)) {
+    segment = parts[i];
+    if (!(segment in node)) {
       return false;
     } else {
-      if (i + 1 === parts.length)
-        return node[parts[i]].methods;
+      if (i + 1 === parts.length) {
+        return node[segment].methods;
+      }
     }
-    node = node[parts[i]];
+    node = node[segment];
   }
 };
 
@@ -43,21 +52,23 @@ function getPathParts(path) {
   path = path.trim();
   var chunks = path.split('/');
   var parts = [];
-  chunks.forEach(function(chunk) {
-    if (chunk) {
-      parts.push(chunk);
+
+  for (var i in chunks) {
+    if (chunks[i]) {
+      parts.push(chunks[i]);
     }
-  });
+  }
   return parts;
 }
 
-function addVerbHandler(node, verb, handler) {
+function addVerbHandler(node, route, handler) {
   node.methods = node.methods || {};
+  var verb = route.verb;
   if (node.methods[verb]) {
     console.warn(
-      'WARN: A handler was already registered for %s /api%s : Ignoring the new handler-%s',
-      verb.toUpperCase(),
-      handler.path || '/[unknown]',
+      'WARN: A handler was already registered for %s /api%s : Ignoring the new handler %s',
+      route.verb.toUpperCase(),
+      route.fullPath || '/[unknown]',
       handler.methodName || '<unknown>'
       );
     return;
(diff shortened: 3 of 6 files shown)


lib/rest-adapter.js
```diff
@@ -403,6 +411,14 @@ RestAdapter.prototype._registerMethodRouteHandlers = function(router,
...
   }
+
+  //For now: exclude paths with parameters like :id
+  //and handler functions with err obj i.e. (err, req,res,next)
+  if (!route.path.includes(':') && handler.length < 4) {
+    handler.methodName = sharedMethod.name;
+    handler.path = route.fullPath;
+    customRouter.registerPathAndHandlers({ verb: verb, fullPath: route.fullPath }, handler);
+  }
   router[verb](route.path, handler);
```
gunjpan: This block is causing trouble :( All tests with `npm test` passed on local, CI fails

gunjpan: Moving this line in `else` block causes failures on local. We need to define a clear strategy to port custom router.

lib/rest-adapter.js
```diff
@@ -258,6 +259,10 @@ RestAdapter.prototype.createHandler = function() {
 
   var handleUnknownPaths = this._shouldHandleUnknownPaths();
 
+  // Mount rest-router
+  var customRouter = new RestRouter();
```
bajtos: `custom` is not very descriptive. How about `fastRouter` instead?

lib/rest-adapter.js
```diff
@@ -272,6 +277,8 @@ RestAdapter.prototype.createHandler = function() {
         var sharedMethod = restMethod.sharedMethod;
         debug('    method %s', sharedMethod.stringName);
         restMethod.routes.forEach(function(route) {
+          //add fullPath for Trie insertion
```
bajtos: nitpick: we are adding a space after `//`, i.e. `// add...`

lib/rest-adapter.js
```diff
@@ -403,6 +411,14 @@ RestAdapter.prototype._registerMethodRouteHandlers = function(router,
...
     verb = 'delete';
   }
+
+  //For now: exclude paths with parameters like :id
+  //and handler functions with err obj i.e. (err, req,res,next)
+  if (!route.path.includes(':') && handler.length < 4) {
+    handler.methodName = sharedMethod.name;
+    handler.path = route.fullPath;
+    customRouter.registerPathAndHandlers({ verb: verb, fullPath: route.fullPath }, handler);
+  }
```
bajtos: Please move as much of this code into `RestRouter` as possible.

```js
if (fastRouter.canRegister(route.path, handler)) {
  fastRouter.registerPathAndHandler({
    verb: verb,
    fullPath: route.fullPath,
    methodName: sharedMethod.name
  }, handler); // << perhaps use "stringName" instead?
}
```

I am wondering, can we make the API of our new router compatible with the vanilla Express Router? Instead of `fastRouter.registerPathAndHandlers`, one would call:

```js
// function(path, handler, options)
fastRouter[verb].use(fullPath, handler, { displayName: sharedMethod.stringName });
```

I borrowed "displayName" from here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/displayName. It was initially proposed for ES6, but never made it.

lib/rest-adapter.js
bajtos: I would expect that L422 is not invoked when the method was registered with the new fast router. I'll check.

lib/rest-adapter.js
```diff
@@ -258,6 +259,10 @@ RestAdapter.prototype.createHandler = function() {
 
   var handleUnknownPaths = this._shouldHandleUnknownPaths();
 
+  // Mount rest-router
+  var customRouter = new RestRouter();
+  root.use(customRouter.handle());
```
bajtos: I find this weird, why can't we directly `root.use(customRouter)`? I'd like our new router be as much compatible with the built-in Express router as possible.

bajtos: I see, `root.use` expects a handler function, while our router is an object instead.

You can use the approach used by express ([source](https://github.com/strongloop/express/blob/4fb3653b6ce9bd9080ceadb267f2cffc32413444/lib/router/index.js#L42-L60)), even though it's not very pretty.

```js
// lib/rest-router.js
var RestRouter = module.exports = function(options) {
  var opts = options || {};

  var router = function(req, res, next) {
    router.handle(req, res, next);
  }

  // mixin Router class functions
  router.__proto__ = RestRouter;

  router.trie = new Trie();
  router.options = options; 

  return router;
}
```

lib/rest-router.js
```diff
@@ -0,0 +1,37 @@
+module.exports = RestRouter;
+
+var url = require('url');
+var Trie = require('./trie');
+
+function RestRouter(options) {
+  this.trie = new Trie();
```
bajtos: Do you expect router consumers to access `router.trie` directly? I think this property should be treated as private, in which case it should be called `_trie`.

lib/rest-router.js
```diff
@@ -0,0 +1,37 @@
...
+  this.options = options;
```
bajtos: ditto

lib/rest-router.js
```diff
@@ -0,0 +1,37 @@
...
+  var path = normalizePath(route.fullPath);
+  var trie = this.trie;
+  trie = trie.add(path, route.verb, handler);
+};
+
+//handle request, match path and if found, invoke handler method
+RestRouter.prototype.handle = function() {
+  var trie = this.trie;
+  return function(req, res, next) {
+    var verb = req.method.toLowerCase() || 'all';
```
bajtos: This is weird. The "all" verb can be used when registering a handler. HTTP request always provide `req.method`.

Please add a unit-test to verify how a handler for `all` verb(s) works.

lib/rest-router.js
```diff
@@ -0,0 +1,37 @@
...
+    var path = normalizePath(req.url, this.options);
+    var methods = trie.find(path);
+    if (methods && verb in methods) {
+      methods[verb](req, res, next);
+    } else next();
```
bajtos: Be consistent with braces - either use them in both "if" and "else" branch, or don't use them in both branches.

I think we tend to use braces unless the whole if + statement fits a single line.

```js
if (...) {
  methods[verb](...);
} else {
  next();
}
```

lib/trie.js
```diff
@@ -0,0 +1,66 @@
...
+
+function Trie() {
+  this.methods = {};
+}
+
+//insert new nodes to Trie
+Trie.prototype.add = function(path, verb, handler) {
+  var node = this;
+  var parts = getPathParts(path);
+  if (parts.length) {
```
bajtos: Please use a reverse logic + early return to simplify the body of this function.

```js
if (!parts.length) {
  addVerbHandler(node, verb, handler);
  return this;
}

for (...) {
  ...
}
```

lib/trie.js
```diff
@@ -0,0 +1,66 @@
...
+    for (var i = 0; i < parts.length; i++) {
+      if (!(parts[i] in node)) {
+        node[parts[i]] = new Trie();
+      }
+      if (i + 1 === parts.length) {
+        addVerbHandler(node[parts[i]], verb, handler);
+      }
+      node = node[parts[i]];
+    }
+  } else { // if we would ever encounter this situation?
```
bajtos: Isn't this the case of `add('/', 'GET', handler)`? Please add a unit-test to verify.

lib/trie.js
```diff
@@ -0,0 +1,66 @@
...
+  this.methods = {};
+}
+
+//insert new nodes to Trie
+Trie.prototype.add = function(path, verb, handler) {
+  var node = this;
+  var parts = getPathParts(path);
+  if (parts.length) {
+    for (var i = 0; i < parts.length; i++) {
+      if (!(parts[i] in node)) {
```
bajtos: please extract `parts[i]` into a variable to avoid repetition, e.g. `var segment = parts[i]`. Or rename `parts` to `pathParts` and then use `var part = pathParts[i]`.

lib/trie.js
```diff
@@ -0,0 +1,66 @@
...
+  }
+  return this;
+};
+
+//Look for matching path and return handler method for a match
+Trie.prototype.find = function(path) {
+  var node = this;
+  var parts = getPathParts(path);
+  for (var i = 0; i < parts.length; i++) {
+    if (!(parts[i] in node)) {
```
bajtos: ditto - extract `parts[i]` into a variable.

lib/trie.js
```diff
@@ -0,0 +1,66 @@
...
+      node = node[parts[i]];
+    }
+  } else { // if we would ever encounter this situation?
+    addVerbHandler(node, verb, handler);
+  }
+  return this;
+};
+
+//Look for matching path and return handler method for a match
+Trie.prototype.find = function(path) {
```
bajtos: Since `find` is accepting both path+verb, it's better to modify `find` to accept both path+verb too, for the sake of consistency.

lib/trie.js
```diff
@@ -0,0 +1,66 @@
...
+
+function getPathParts(path) {
+  path = path.trim();
+  var chunks = path.split('/');
+  var parts = [];
+  chunks.forEach(function(chunk) {
+    if (chunk) {
+      parts.push(chunk);
+    }
+  });
```
bajtos: This can be simplified to `var parts = chunks.filter(function(c) { return !!c; })`.

However, we have learned that `forEach` (or `filter`) are better avoided in code paths frequently invoked, as they incur performance overhead due to the need to allocate a closure for the callback function.

Please rewrite this method to use `for` instead.

lib/trie.js
```diff
@@ -0,0 +1,66 @@
...
+    addVerbHandler(node, verb, handler);
+  }
+  return this;
+};
+
+//Look for matching path and return handler method for a match
+Trie.prototype.find = function(path) {
+  var node = this;
+  var parts = getPathParts(path);
+  for (var i = 0; i < parts.length; i++) {
```
bajtos: btw this can be simplified to `for (var i in parts)`

lib/trie.js
```diff
@@ -0,0 +1,66 @@
...
+    }
+  });
+  return parts;
+}
+
+function addVerbHandler(node, verb, handler) {
+  node.methods = node.methods || {};
+  if (node.methods[verb]) {
+    console.warn(
+      'WARN: A handler was already registered for %s /api%s : Ignoring the new handler-%s',
```
bajtos: > handler-%s

Why the dash character? I would simply use a space: `... handler %`.

lib/trie.js
```diff
@@ -0,0 +1,66 @@
...
+      verb.toUpperCase(),
+      handler.path || '/[unknown]',
+      handler.methodName || '<unknown>'
```
bajtos: Storing metadata on the handler is not great, can you pass it in `addVerbHandler` arguments instead?

test/rest-router.test.js
```diff
@@ -0,0 +1,60 @@
+var assert = require('assert');
+var expect = require('chai').expect;
+var RestRouter = require('../lib/rest-router');
+
+describe('Custom Router', function() {
```
bajtos: Please rename to match the class name from `../lib/rest-router`

test/rest-router.test.js
```diff
@@ -0,0 +1,60 @@
...
+  var restRouter;
+  var route = { verb: 'get', fullPath: '/Planets' };
+  var handler = function(req, res, next) { res.handled = true; };
+  var req = {
+    method: 'get',
+    url: '/Planets',
+  };
+  var res = {};
```
bajtos: Mocking out req/res objects is a bad idea, Express maintainers are actively discouraging people from doing that.

Please create a regular express app to test the router. Also keep in mind that you need to write tests for both values of `options.caseSensitive`.

```js
var app;
beforeEach(function() {
  app = expres();
});

it('invokes the handler function for matching verb+path', function(done) { 
  var router = new RestRouter(); // optionally set options here
  app.use(router);
  router.get('/Planets', function(req, res, next) { res.end('hello'); });
  executeRouter(app, 'get', '/Planets', function(err, response, body) {
    expect(body).to.equal('hello');
  });
});

function executeRouter(app, verb, urlPath, callback) {
  var server = http.createServer(function(req, res) {
    app.handle(req, res, callback);
  });

  supertest(server)[verb](urlPath).end(callback);
}
```

This is just a mockup based on https://github.com/strongloop/loopback/blob/9a33602458712afecedd25062f02bd8b5c4e01a5/test/app.test.js#L959-L974, it needs a bit of polishing. I think ultimately we would like to arrive at something like this:

```js
beforeEach(function(done) {
  // setup the app
  // setup var request = supertest(app or server)
});

it('does something', function(done) {
  // setup the router
  // register handlers (routes)
  request.get('/path').expect(200).expect('expected-body').end(done);
});
```

test/rest-router.test.js
```diff
@@ -0,0 +1,60 @@
...
+    });
+  });
+
+  describe('restRouter.handle()', function() {
+    it('returns a function --which invokes matching handler-- with three arguments',
+      function() {
+        restRouter.registerPathAndHandlers(route, handler);
+        var returned = restRouter.handle();
+        expect(returned).to.have.length(3);
+      });
```
bajtos: Instead of asserting on function API, it's better to invoke the handler via express as I described above. That way we verify that the router actually works with express, instead of verifying that it matches an API that we think will work.

bajtos:
@gunjpan please review my comments in our earlier conversations (e.g. in the Spike/POC pull request) and add all unit-tests that I mentioned there. It's ok to add them as "it.skip" initially, I just want to make sure we don't forget about any of them.

lib/rest-adapter.js
bajtos: Here is a clue - I added some logging to `RestRouter`. 

```
$ mocha -g "strong-remoting-rest call of constructor method should allow arguments in the query"

  strong-remoting-rest
    call of constructor method
REGISTER all /testclass
HANDLE get /testclass/
      1) should allow arguments in the query
```

Please add `debug()` logs to both `registerPathAndHandlers` and `handle` so that users can easily get logs like that without any code changes. Take a look at how `express.Router` logs debug messages and add similar logs to your new router.

bajtos: So the thing is, the handler is registered for *all* verbs on `/testclass`. The request for `get /testclass` does not recognise this handler.

lib/trie.js
gunjpan: @bajtos: yes. I purpose-fully put the comment for your views. 
The thing is, every handler, in our `rest-adapter.js` is related to one restClass, thus it will always have model name in the `fullPath`. So, was not sure if we should handle this case. 
TODO: Will remove the comment and add a unit-test

lib/rest-adapter.js
gunjpan: @bajtos: In our case, when we have one handler for verb: `all` and another handler for verb `get`, our router will register both, and will invoke `get` one on `GET` request. 
So, the question is: do we drop `get` handlers when `all` is present or chain them such as:  `all:handler` executes -> `get:handler` executes (if this' the case, then we'll have multiple, at least two, handlers to execute)

please advise. Thanks.

lib/trie.js
bajtos: When it is easy to do so, (like in this case), then we should try to make this new router usable for general audience. If the benchmarks results are good, then we may want to extract this fast router to a standalone package for everybody to use.

lib/rest-adapter.js
bajtos: @gunjpan I'll answer with the same question as before. What does the built-in router do in this case?

bajtos:
@gunjpan on more task from our earlier conversations: please add the test suite that will run the same code (tests) against both `express.Router` and `RestRouter`. The sooner you do that, the sooner we can start adding tests for edge cases like https://github.com/strongloop/strong-remoting/pull/282#discussion_r51957509.

lib/trie.js
gunjpan: @bajtos: In our implementation, `find` is only accepting the `path` as it's being called by `RestRouter.handle()`. `find` returns a `methods` object which has all `{verb:handlers}` registered for that path. 
I'm using `RestRouter.handle()` to do the matching work for `verb`. This way, we can keep our matching criteria for `verb+path` (such as, we may need to handle multiple handler for same verb+path) separated from trie implementation.

For the sake of consistency, I've changed the `trie.add(path, verb, handler)` to `trie.add(route, handler)`
and have kept `trie.find(path)` as it is.

lib/rest-adapter.js
gunjpan: @bajtos : the built-in router places `all` verb handler on the top of the handler stack, and keeps a flag to check if the `path` has an `all` handler, if yes, it invokes the `all` handler first and runs other registered handlers
In our case, `trie` has a placeholder for `all` handler, and we can check, but we are only inovking single handler. 
I'm thinking of changing it in a way that: 
in `all: handler(req,res,next)` , `next` points to the second handler in the stack for the path and so on.. I will commit this approach in the next commit.

Or, should you have any better idea for this scenario -- after looking at recent commit -- please advise. Thank you :)

lib/rest-adapter.js
```diff
@@ -403,6 +411,11 @@ RestAdapter.prototype._registerMethodRouteHandlers = function(router,
     // Express 4.x only supports delete
     verb = 'delete';
   }
+
+  if (fastRouter.canRegister(route, handler)) {
+    handler.methodName = sharedMethod.name;
+    fastRouter[verb](route, handler);
```
bajtos: I think adding extra properties on the `handler` method makes it more difficult to understand what's going on, because the other parts of the code are depending on the fact that `fn.methodName` is defined, which is not obvious.

That's why I proposed an API that is an extension of the original Express API:

```js
// express:
router[verb](route, handler)

// proposed extension:
router[verb](route, handler, options) 
```

OTOH, this is a minor detail. Please considered my comment, and if you still feel that `handler.methodName` is better, then let's do it your way.

lib/rest-router.js
```diff
@@ -0,0 +1,66 @@
...
+
+  router.__proto__ = RestRouter;
+  router._trie = new Trie();
+  router.options = opts;
+
+  return router;
+};
+
+// express style API for registering route handlers: ex. route[method](path, handler)
+// add verb:hanlder for path to trie DS
```
bajtos: typo: `verb:hanlder`

lib/rest-router.js
```diff
@@ -0,0 +1,66 @@
+
+var url = require('url');
+var http = require('http');
+var Trie = require('./trie');
+
+var methods = getCurrentNodeMethods();
```
bajtos: since `methods` should be treated as a constant, please use all-capital letters in the name. I would add `methods` to the name, to make the distinction between HTTP methods and shared/remote methods clear.

```js
var HTTP_METHODS = getHttpMethods();
```

lib/rest-router.js
```diff
@@ -0,0 +1,66 @@
...
+  path = url.parse(path).pathname;
+  if (!options || !(options.caseSensitive)) {
+    path = path.toLowerCase();
+  }
+  return path;
+}
+
+// get list of all HTTP methods
+function getCurrentNodeMethods() {
+  return http.METHODS && http.METHODS.map(function lowerCaseMethod(method) {
```
bajtos: This does not work in older versions of Node (e.g. v0.10) where `http.METHODS` is not defined. Please fix.

lib/trie.js
```diff
@@ -0,0 +1,77 @@
...
+}
+
+function addVerbHandler(node, route, handler) {
+  node.methods = node.methods || {};
+  var verb = route.verb;
+  if (node.methods[verb]) {
+    console.warn(
+      'WARN: A handler was already registered for %s /api%s : Ignoring the new handler %s',
+      route.verb.toUpperCase(),
+      route.fullPath || '/[unknown]',
```
bajtos: I'd personally use `[unknown path]` instead of `/[unknown]` - I think the former makes it more clear what is unknown.

lib/trie.js
```diff
@@ -0,0 +1,77 @@
...
+      handler.methodName || '<unknown>'
```
bajtos: I'd use `<anonymous>` as the default method name, that's what V8 uses.

```
> (function() { throw new Error(); })()
Error
    at repl:1:22
    at repl:1:37
    at REPLServer.self.eval (repl.js:110:21)
    at Interface.<anonymous> (repl.js:239:12)
    at Interface.emit (events.js:95:17)
    at Interface._onLine (readline.js:203:10)
    at Interface._line (readline.js:532:8)
    at Interface._ttyWrite (readline.js:761:14)
    at ReadStream.onkeypress (readline.js:100:10)
    at ReadStream.emit (events.js:98:17)
```

lib/rest-adapter.js
bajtos: If you decide to keep the name on the `handler` function, then please choose a more descriptive property name. The problem with `methodName` is that it's not clear whether it refers to HTTP method or shared/remote method.

test/trie.test.js
```diff
@@ -0,0 +1,81 @@
...
+      trie.add(setRoute(route, '/Planets/count', 'get'), 'getCount()');
+      trie.add(setRoute(route,'/Planets/count', 'post'), 'postCount()');
+      var handlers = trie.find('/Planets/count');
+      expect(handlers.get).to.be.equal('getCount()');
+      expect(handlers.post).to.be.equal('postCount()');
+    });
+  });
+});
+
+function setRoute(route, fullPath, verb) {
```
bajtos: I find this a bit weird. Why can't you simply use a factory function?

```js
function aRoute(verb, fullPath) {
  return {
    verb: verb,
    fullPath: fullPath
  };
});
```

lib/trie.js
```diff
@@ -0,0 +1,77 @@
+
+module.exports = Trie;
+
+function Trie() {
+  this.methods = {};
+}
+
+// insert new nodes to Trie
+Trie.prototype.add = function(route, handler) {
```
bajtos: I'd personally use `verb` + `fullPath` arguments instead of `route`, e.g.

```js
Trie.prototype.add = function(verb, fullPath, handler) {
  // ...
};
```

That way the `Trie` is not coupled with strong-remoting's routes and also it's more obvious what data is needed for `add`.

lib/trie.js
```diff
@@ -0,0 +1,77 @@
...
+      addVerbHandler(node[segment], route, handler);
+    }
+    node = node[segment];
+  }
+
+  return this;
+};
+
+// Look for matching path and return methods for a match
+Trie.prototype.find = function(path) {
```
bajtos: IIRC I was commenting earlier that I would like this to become `Trie.prototype.find = function(path, verb)`.

lib/rest-adapter.js
bajtos: I could not find this comment on GH, so I am responding here.
> the built-in router places all verb handler on the top of the handler stack, and keeps a flag to check if the path has an all handler, if yes, it > invokes the all handler first and runs other registered handlers> In our case, trie has a placeholder for all handler, and we can check, but we are only inovking single handler.
That makes perfect sense to me, now we just need to improve error logging so that the user understands what's happening under the hood.
 
1) If there is an "all" handler already registered for a path, and users is attempting to add another handler, then we should do the same thing we do now when the second handler is registered for the same "path+verb"
 
2) When registering an "all" handler, check if there are any existing verb-specific handlers for the same path. If there are, then report a warning for each of them, since they won't be invoked.
 
BTW are you sure the "all" verb handler is on the top? I would expect it's at the bottom, i.e. more specific handlers (GET, POST) are run before generic ones (ALL).
> I'm thinking of changing it in a way that:> in all: handler(req,res,next) , next points to the second handler in the stack for the path and so on.. I will commit this approach in the next commit.> Or, should you have any better idea for this scenario -- after looking at recent commit -- please advise. Thank you :)
AFAIK, all strong-remoting handler functions either send back the response (and never call next), or fail with an error and call next(err). In both cases the flow control skips all subsequent handler functions.
 
This is why we keep only the first handler for each path+verb (instead of an array), and that's why I believe we don't need to chain handlers for "all" verb either. Am I missing something?
 
Miroslav
 
 
----- Original message -----From: Gunjan Pandya <notifications@github.com>To: strongloop/strong-remoting <strong-remoting@noreply.github.com>Cc: Miroslav Bajtos/Czech Republic/IBM@IBMCZSubject: Re: [strong-remoting] Implement a faster REST router (#282)Date: Mon, Feb 8, 2016 9:48 AM 
In lib/rest-adapter.js:
> @@ -403,6 +411,14 @@ RestAdapter.prototype._registerMethodRouteHandlers = function(router,>      // Express 4.x only supports delete>      verb = 'delete';>    }> +> +  //For now: exclude paths with parameters like :id> +  //and handler functions with err obj i.e. (err, req,res,next)> +  if (!route.path.includes(':') && handler.length < 4) {> +    handler.methodName = sharedMethod.name;> +    handler.path = route.fullPath;> +    customRouter.registerPathAndHandlers({ verb: verb, fullPath: route.fullPath }, handler);> +  }>    router[verb](route.path, handler);
@bajtos : the built-in router places all verb handler on the top of the handler stack, and keeps a flag to check if the path has an all handler, if yes, it invokes the all handler first and runs other registered handlersIn our case, trie has a placeholder for all handler, and we can check, but we are only inovking single handler.I'm thinking of changing it in a way that:in all: handler(req,res,next) , next points to the second handler in the stack for the path and so on.. I will commit this approach in the next commit.
Or, should you have any better idea for this scenario -- after looking at recent commit -- please advise. Thank you :)

lib/rest-adapter.js
gunjpan: @bajtos : I will set it to `sharedMethod.stringName` thus it would be clear if it's a prototype method or a static method, and as you recommended, will use `displayName` as:
`handler.displayName = sharedMethod.stringName`
Is that fine?

lib/rest-router.js
gunjpan: @bajtos : Good catch ;)

lib/rest-router.js
gunjpan: @bajtos: I borrowed that part from a module: [npm](https://www.npmjs.com/package/methods) which is used by Express. I didn't want to add any further package dependencies.
Your thoughts on using it, please

lib/trie.js
gunjpan: @bajtos : yes, and I commented on an outdated difference [here](https://github.com/strongloop/strong-remoting/pull/282#discussion-diff-51856025R13).

lib/rest-router.js
bajtos: >  I didn't want to add any further package dependencies.

No need to reinvent a wheel here, it's ok to add `methods` as a new dependency.

lib/rest-adapter.js
gunjpan: > BTW are you sure the "all" verb handler is on the top? I would expect it's at the bottom, i.e. more specific handlers (GET, POST) are run before generic ones (ALL).
Yes, I'm pretty sure about it from express API docs:
http://expressjs.com/en/4x/api.html#router.all

and it's also the matter of registering the `router.all()` first to execute it first in middleware chain.

bajtos: As I understand docs, `all` is treated the same way as other, regular verbs. If you put your route handler as the first one, it will work as you describe, but if you put it as the last one, it will work as I described it. So the task for you is to find how strong-remoting is dealing with this. IIRC there is a custom sort of routes, does it take verbs into account too?

gunjpan: @bajtos : strong-remoting sorts routes in the following custom way: It first sorts by `verb` and then sorts path-part by part.

bajtos: Oh, I see. And because "all" starts with the first letter in the alphabet, it will end up being registered as the first one.
 
To be honest, this looks like a bug to me, but I am not sure if it's a good idea to fix it now, when we are possibly introducing other breaking changes caused by the new router implementation.
 
Could you please leave a comment in the code, explaining why we are treating "all" verb handlers as the first ones and open a GH issue to track the task of changing this to the more sensible approach where "all" becomes a fall-back only? The code comment should link to this new issue too.
 
Thanks,
Miroslav

richardpringle:
I want to propose a different name for `Trie` and `trie.js`. Maybe something a little more descriptive like `RouteTrie`, thoughts?

When I see `Trie` I immediately assume that the class is simply the trie data structure and is not necessarily related to routing.

bajtos:
> Maybe something a little more descriptive like `RouteTrie`, thoughts?

It may be bettero to treat the fact that we use Trie as an implementation detail and don't leak it in the name at all. Then we can call the class `RouteTable` or `RouterLookup` or something along those lines. Thoughts?

richardpringle:
@bajtos, is there any reason not to be transparent? Someone from the community might see it and have a better idea... 

I'm not picky about that detail though, I like `RouteTable` as well.

bajtos:
cc @raymondfeng PTAL too

bajtos:
>  is there any reason not to be transparent? Someone from the community might see it and have a better idea... 

So what will happen if we decide to change the implementation later in the future and use a different data structure? Will we have a `RouteTrie` that's implemented using [red-black tree](https://en.wikipedia.org/wiki/Red%E2%80%93black_tree) (for example)? Will we rework all code that is using (calling) this class, even though there was no API change beyond the name?

That's why I am proposing to use a name that describes how the class is intended to be used (to look up routes), not how it is implemented (via Trie now). The former should stay the same even in case of changes in the future, the latter will not stay the same.

richardpringle:
@bajtos, this is still very much a work in progress and there are still a few changes that need to be made that you had originally recommended for @gunjpan, but I would like you to please review before I continue to work on this.

I would also like your opinion on any more test cases that should be added.

richardpringle:
@gunjpan @davidcheung, could you also please take a look.

lib/rest-adapter.js
```diff
@@ -425,6 +433,11 @@ RestAdapter.prototype._registerMethodRouteHandlers = function(router,
     // Express 4.x only supports delete
     verb = 'delete';
   }
+
+  // if (fastRouter.canRegister(route, handler)) {
```
bajtos: I don't think it's reasonable to expect that fastRouter will support all different route flavours supported by express (e.g. regular expressions). That's why it's important to call `canRegister` (or perhaps `canHandle` is a better name?) here.

lib/rest-router.js
```diff
@@ -0,0 +1,59 @@
...
+
+  if (methods && verb in methods) {
+    methods[verb](req, res, next);
+  } else {
+    next();
+  }
+};
+
+// For now: exclude paths with parameters like :id
+// and handler functions with err obj i.e. (err, req, res, next)
```
bajtos: Is this comment still relevant?

lib/rest-router.js
```diff
@@ -0,0 +1,59 @@
...
+
+// RestRouter.canRegister = function(route, handler) {
```
bajtos: As mentioned earlier, we should still preserve this method.

test/rest-router.test.js
```diff
@@ -0,0 +1,93 @@
...
+    // this doesn't make sense to me that I have to call router.get(...)
+    // as well as have route.verb = get
+    // it doesn't work without it
+    router.get(route, function(req, res, next) {
+      console.log('hit');
+      res.end(id);
+    });
+    app.use(router);
+    console.log(router._trie);
+    supertest(app)
```
bajtos: Please simplify the test setup to avoid repetition of the boiler plate setting up an app with a fast router.

IIRC, each call of `supertest(app)` creates a new http server and binds to a new port, which slows down tests especially on Windows. A better solution is to create a single http server and share it by all tests, see https://github.com/strongloop/strong-remoting/blob/bcc2ad7b08fd37e9264edda2aab3526f398169f0/test/rest-coercion.test.js#L30-L46

test/rest-router.test.js
```diff
@@ -0,0 +1,93 @@
...
+
+  // test handle param
+  it('handles a request with a parameter', function(done) {
+    var app = express();
+    var router = new RestRouter();
+    var route = { fullPath: '/example/:id', verb: 'get' };
+    var id = '1';
+    // this doesn't make sense to me that I have to call router.get(...)
+    // as well as have route.verb = get
+    // it doesn't work without it
```
bajtos: This should be fixed.

test/rest-router.test.js
```diff
@@ -0,0 +1,93 @@
...
+    router.get(route, function(req, res, next) {
+      console.log('hit');
+      res.end(id);
```
bajtos: This test always passes, because you always return `id` from the variable. You need to return `req.params.id` instead.

bajtos: In TDD, it's crucial to start with a failing unit test, to see the test fail and fail for the right (expected) reason. Without that, it's easy to write a test that always passed (like this one) or a test that fails for completely different reasons because it's based on wrong assumptions about how the code under test works.

Last but not least, seeing the test fail is the opportunity to verify that the failure message makes it clear what went wrong, what was expected and actual result.

bajtos:
> this is still very much a work in progress and there are still a few changes that need to be made that you had originally recommended for @gunjpan, but I would like you to please review before I continue to work on this.

I am not sure what exactly to look out for during review. I did a quick check to see if there is anything obvious and left few comments above.

>  I would also like your opinion on any more test cases that should be added.

For me, it's most important to have an integration test suite to verify that the new fast router behaves exactly same as the default express router.

I think the best approach is to create a parameterized suite that will be run against the new fastRouter and against the Express router. For example, something along these lines:

```js
describe('fast router', function() {
  testRouter(FastRouter);
});

describe('express Router', function() {
  testRouter(express.Router);
});

function testRouter(Router) {
  var router, request;
  before(setupHttpServer);
  after(stopHttpServer);

  // individual tests calling the provided `router` under the hood
  it('...', function(done) {
    router.get('/', function(req, res, next) {
      // ...
    });
    request.get('/')
      .expect(..., done);
  });

  function setupHttpServer(done) {
    router = new Router();
    // create http server handling requests via router
    // create supertest instance invoking this server
  }
}
```

As for test cases, we should test all HTTP verbs, then different URL paths (`/`, path with one segment, path with multiple segments, path with a trailing `/`, check how double/triple slash is handled - at the start `//api`, in the middle `/api//cars`, and at the end `/api/cars//`). Check how `all` is handled - what is the priority compared to specific verbs like `get`/`post`. Look at express source code dealing with path parameters and add test cases to verify we correctly handle edge cases.

bajtos:
We should also verify that when you register a handler for a given verb+path, other endpoints will not trigger the handler.

Verify the priority of paths with path parameters vs. paths with static names, think about `GET /api/cars/:id` vs. `GET /api/cars/findOne`: `/api/cars/1` should be routed to `findById()`, `/api/cars/findOne` should be routed to `findOne()`.

lib/rest-adapter.js
richardpringle: Did you look at the code for `canRegister`? It only registers with express if the handler function has more than 3 arguments or if there is a ':' in the path (in other words, any route with a parameter).

So I just want to be clear and say that you are not suggesting to use the function that exists but instead to make sure that the router *can handle* the route, otherwise pass it to the express router.

Note: this should be done with an `if else` statement instead of just an `if`.

lib/rest-router.js
richardpringle: I believe the 2nd part is

lib/rest-router.js
```diff
@@ -0,0 +1,59 @@
...
+  var methods = trie.find(path);
+
+  if (methods && verb in methods) {
+    methods[verb](req, res, next);
+  } else {
+    next();
+  }
+};
+
+// For now: exclude paths with parameters like :id
```
richardpringle: ^ remove this

lib/rest-router.js
richardpringle: We need to re-write this method.

test/rest-router.test.js
richardpringle: Sounds good. I never planned on leaving the repeated code in each test. I wanted to make sure I was doing this right before I removed it.

test/rest-router.test.js
richardpringle: I understand how TDD works. The test was originally failing because `GET /example/1` was not a registered route, `/:id` was not being replaced by the parameter. I was not testing to see if `req.params.id` was being populated properly. 

I do however agree that it makes sense to test this here as well.

richardpringle:
@bajtos 
>I am not sure what exactly to look out for

I wanted you to look out for the things that you found. It's better to catch those mistakes early instead of having to do a big re-factor at the end. It's a much faster process to review as I move forward instead of one big repeating review/re-factor cycle at the end.

richardpringle:
Also, I keep getting the same failure for multiple tests on Node 0.10 and 0.12
```shell
     TypeError: undefined is not a function
      at Trie.find (lib/trie.js:9:2363)
      at Function.RestRouter.handle (lib/rest-router.js:9:1870)
      at router (lib/rest-router.js:9:688)
      at corsHandler (lib/rest-adapter.js:9:13002)
      at lib/rest-adapter.js:9:11460
      at test/phase-handlers.test.js:20:30
      at HTTPParser.parserOnIncoming [as onIncoming] (_http_server.js:492:12)
      at HTTPParser.parserOnHeadersComplete (_http_common.js:111:23)
      at Socket.socketOnData (_http_server.js:343:22)
      at readableAddChunk (_stream_readable.js:163:16)
      at Socket.Readable.push (_stream_readable.js:126:10)
      at TCP.onread (net.js:540:20)
```

bajtos:
Few more things to test: values of path parameters should be url-decoded, i.e. `/api/cars/a+b` should produce value 'a b', similarly for `%` based sequences.

lib/rest-adapter.js
bajtos: > So I just want to be clear and say that you are not suggesting to use the function that exists but instead to make sure that the router can handle the route, otherwise pass it to the express router.

Correct.

> Note: this should be done with an `if else` statement instead of just an `if`.

:+1:

lib/rest-router.js
bajtos: You commented-out  `if (router.canRegister`)` , therefore no paths/handlers are excluded now AFAICT. I find this comment confusing, because it describes situation before you made your changes and seems to be in contradiction with the current version of the code.

lib/rest-router.js
bajtos: +1

bajtos:
> Also, I keep getting the same failure for multiple tests on Node 0.10 and 0.12

That's weird, because `lib/trie.js:9` is this now:

```js
Trie.prototype.add = function(route, handler) {
```

@richardpringle can you reproduce the problem locally on your machine too? (I personally use and recommend [nvm](https://github.com/creationix/nvm) for quick & easy switching between node versions.)

lib/trie.js
```diff
@@ -0,0 +1,70 @@
...
+  var node = this;
+  var parts = getPathParts(path);
+  var segment;
+
+  for (var i = 0; i < parts.length; i++) {
+    segment = parts[i];
+    if (!(segment in node) && !node.param) return false;
+    if (node.param) {
+      segment = Object.keys(node).find(function(key) {
```
richardpringle: `array.prototype.find` is not an ES5 function

test/trie.test.js
richardpringle: The above is irrelevant if you pass in `verb, `path`, and `handler` to necessary functions as suggested [in this comment](https://github.com/strongloop/strong-remoting/pull/282/files#r52163617)

bajtos:
@gunjpan I am re-assigning this back to you. Please ping me when there patch is ready for another round of review.

gunjpan:
@bajtos : Thank you. I may need to spend some time to see what richard did, rebase and see which way to go. Will update you when done. Thanks.

richardpringle:
@gunjpan, let me know if you need any help!

bajtos:
I am closing this patch for now, we can re-open it when the time to work on this feature comes again.

Issue #35, opened 2014-01-10
Stream as an accepts/returns argument


raymondfeng:
I'm trying to implement a 'download' remote method that will pipe a stream into http response. For example,

function download(..., cb) {
     var in = fs.createReadStream('1.jpg', ...);
     in.pipe(res);
}

There are a few options:

1. Have the download function to callback with a ReadStream so that strong-remoting will pipe it into the response.

2. Invent a way to declare req/res parameters so that strong-remoting can inject them for the method invocation.

3. Make HTTP context accessible by calling Context.getCurrent().

bajtos:
As of #218, each "returns" argument can be configured to modify response status code or response header.

bajtos:
When implementing this feature, we should take into account swagger type "file". It would be nice if the implementation in strong-remoting allowed us to support files in loopback-explorer. See https://github.com/strongloop/loopback-explorer/issues/103

bajtos:
Closing as a duplicate of https://github.com/strongloop/loopback/issues/213.

PR #284, opened 2016-02-24
Add support for "file" return args

bajtos:
1) Fix handling of return args with `http.target:header`.  Arguments mapped to a response http header should not be included in the response body.

2) Allow users to specify `{ type: 'file', root: true }` for the single argument that will be sent directly as the response body.

The following values are supported for file the file argument:
 - String
 - Buffer
 - ReadableStream (anything that exposes .pipe() method)

Example usage:

```js
MyModel.download = function(cb) {
  // this can be for example http.request()
  getTheStreamBody(function(err, stream) {
    if (err) return cb(err);
    // stream can be any of: string, buffer, ReadableStream (e.g. http.IncomingMessage)
    cb(null, stream, 'application/octet-stream');
  });
}

MyModel.remoteMethod('download', {
  isStatic: true,
  returns: [
    { arg: 'body', type: 'file', root: true },
    { arg: 'Content-Type', type: 'string', http: { target: 'header' } },
  ],
});
```

Fix strongloop/loopback#2063
Connect to #35 

/to @ritch please review
/cc @raymondfeng @STRML

### DIFF
--- a/lib/http-context.js
+++ b/lib/http-context.js
@@ -15,6 +15,8 @@ var assert = require('assert');
 var ContextBase = require('./context-base');
 var Dynamic = require('./dynamic');
 var js2xmlparser = require('js2xmlparser');
+var SharedMethod = require('./shared-method');
+
 var DEFAULT_SUPPORTED_TYPES = [
     'application/json', 'application/javascript', 'application/xml',
     'text/javascript', 'text/xml',
@@ -405,6 +407,9 @@ HttpContext.prototype.setReturnArgByName = function(name, value) {
   }
 
   if (returnDesc.root) {
+    // TODO(bajtos) call SharedMethod's convertToBasicRemotingType here?
+    this.resultType = typeof returnDesc.type === 'string' ?
+      returnDesc.type.toLowerCase() : returnDesc.type;
     return;
   }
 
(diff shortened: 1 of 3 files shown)


bajtos:
I have simplified and reworked this feature. Instead of adding a new flag, I added a new type called "file":

```js
{
  returns: { arg: 'body', type: 'file', root: true }
}
```

This way the feature is focused on developer's intention (I want to return a raw file response) instead of implementation details (I want to skip json/xml encoding and send the value in raw form). This should make it easier for consumers of remoting metadata (e.g. loopback-swagger) to correctly handle this case.

ritch:
We should add a note to these docs: http://apidocs.strongloop.com/strong-remoting/#sharedmethod about the new `file` type.

Otherwise LGTM :+1:

bajtos:
> We should add a note to these docs: http://apidocs.strongloop.com/strong-remoting/#sharedmethod about the new file type.

Added, see 0d4ea21

I'd like to document this feature in more details as part of #35, I'll probably create a new "doc" story for that.

raymondfeng:
It's a great feature. What about `accepts`? If one of the args is `file`, we should only handle the content-type `multipart/form-data`. See http://swagger.io/specification/

bajtos:
> It's a great feature. What about `accepts`? If one of the args is `file`, we should only handle the content-type `multipart/form-data`. See http://swagger.io/specification/

@raymondfeng sounds good, please open a new issue to keep track of this feature request.

PR #285, opened 2016-02-26
Add support for "file" return args [2.x]

bajtos:
A back-port of #284 to 2.x

Fix strongloop/loopback#2063


bajtos:
Released as `strong-remoting@2.26.0`.

PR #289, opened 2016-04-06
Handle array of errors.

richardpringle:
Temporary fix for strongloop/loopback#2123

Connect to strongloop/loopback#2123

### DIFF
--- a/lib/rest-adapter.js
+++ b/lib/rest-adapter.js
@@ -369,22 +369,40 @@ RestAdapter.errorHandler = function(options) {
         res.statusCode = err.statusCode || err.status || 500;
       }
 
-      debug('Error in %s %s: %s', req.method, req.url, err.stack);
-      var data = {
-        name: err.name,
-        status: res.statusCode,
-        message: err.message || 'An unknown error occurred',
-      };
+      if (Array.isArray(err)) {
+        var details = err.map(function(it) {
+          var data = generateResponseError(it);
+          delete data.statusCode;
+          return data;
+        });
 
-      for (var prop in err) {
-        data[prop] = err[prop];
+        var msg = 'Failed with multiple errors, see `details` for more information.';
+        err = new Error(msg);
+        err.details = details;
       }
 
-      data.stack = err.stack;
-      if (process.env.NODE_ENV === 'production' || options.disableStackTrace) {
-        delete data.stack;
+      res.send({ error: generateResponseError(err) });
+
+      function generateResponseError(error) {
+        debug('Error in %s %s: %s', req.method, req.url, error.stack);
+
+        var data = {
+          name: error.name,
+          status: res.statusCode,
+          message: error.message || 'An unknown error occurred',
+        };
+
+        for (var prop in error) {
+          data[prop] = error[prop];
+        }
+
+        data.stack = error.stack;
+        if (process.env.NODE_ENV === 'production' || options.disableStackTrace) {
+          delete data.stack;
+        }
+
+        return data;
       }
-      res.send({ error: data });
     }
   };
 };
--- a/test/rest.test.js
+++ b/test/rest.test.js
@@ -1571,6 +1571,38 @@ describe('strong-remoting-rest', function() {
           .end(expectErrorResponseContaining({ message: 'an error' }, done));
       });
 
+      it('should return 500 if an array of errors is thrown', function(done) {
+        var testError = new Error('expected test error');
+        var errArray = [testError, testError];
+
+        function method(error) {
+          return givenSharedStaticMethod(function(cb) {
+            cb(error);
+          });
+        }
+
+        request(app).get(method(testError).url)
+          .set('Accept', 'application/json')
+          .expect(500)
+          .end(function(err, res) {
+            if (err) return done(err);
+            expectedDetail = res.body.error;
+            delete expectedDetail.statusCode;
+
+            request(app).get(method(errArray).url)
+              .set('Accept', 'application/json')
+              .expect(500)
+              .end(function(err, res) {
+                if (err) return done(err);
+                var error = res.body.error;
+                expect(error).to.have.property('message').that.match(/multiple errors/);
+                expect(error).to.include.keys('details');
+                expect(error.details).to.include(expectedDetail);
+                done();
+              });
+          });
+      });
+
       it('should return 500 if an error string is thrown', function(done) {
         remotes.shouldThrow = {
           bar: function(fn) {


richardpringle:
@bajtos, the test case here is a little weird, please advise.

lib/rest-adapter.js
```diff
@@ -369,22 +369,41 @@ RestAdapter.errorHandler = function(options) {
...
+        });
+
+        err = new Error('Failed with multiple errors, see `details` for more information.');
+        err.details = details;
       }
 
-      data.stack = err.stack;
-      if (process.env.NODE_ENV === 'production' || options.disableStackTrace) {
-        delete data.stack;
+      debug('Error in %s %s: %s', req.method, req.url, err.stack);
```
bajtos: This is redundant, because `generateResponseError` already prints a debug message. Please remove.

bajtos: On the second thought, I think it may be a bit surprising that `generateResponseError` prints a debug log as a side-effect. Even you, as an author, got confused.

I am proposing to remove the `debug` statement from `generateResponseError`, and add explicit `debug()` calls in all places where it makes sense. Please check the resulting debug output for both cases (single error, array of errors) to ensure the information printed is useful and easy to understand.

test/rest.test.js
```diff
@@ -1571,6 +1571,41 @@ describe('strong-remoting-rest', function() {
...
+
+        var expectedMessage = 'Failed with multiple errors, see `details` for more information.';
+
+        function method(error) {
+          return givenSharedStaticMethod(function(cb) {
+            cb(error);
+          });
+        }
+
+        request(app).get(method(duplicateError).url)
```
bajtos: I find this confusing. In the context of this request, `duplicateError` is not really duplicated. I am proposing to change it to

```js
var anError = new Error('expected test error');
// or
var testError = new Error('expected test error');
```

test/rest.test.js
```diff
@@ -1571,6 +1571,41 @@ describe('strong-remoting-rest', function() {
...
+            if (err) return done(err);
+            expectedDetail = res.body.error;
+            delete expectedDetail.statusCode;
+
+            request(app).get(method(errArray).url)
+            .set('Accept', 'application/json')
+            .expect(500)
+            .end(function(err, res) {
+              if (err) return done(err);
+              console.log(res.body.error);
```
bajtos: Forgotten console.log?

test/rest.test.js
```diff
@@ -1571,6 +1571,41 @@ describe('strong-remoting-rest', function() {
...
+              var error = res.body.error;
+              expect(error).to.have.property('message', expectedMessage);
```
bajtos: This is brittle. If we decide to reword the message in the future, this assertion will fail.

I think this should be little bit more robust, while it should still fail with the current implementation on master:

```js
expect(error).to.have.property('message').that.match(/multiple errors/);
```

bajtos:
> the test case here is a little weird, please advise.

It looks good to me. What do you find weird? It's possible I am missing something.

richardpringle:
@bajtos 
I just thought there should be a more simple way to verify that the error is correct instead of nesting requests.

lib/rest-adapter.js
richardpringle: @bajtos 
I didn't get confused, I just forgot to delete the line outside of `generateResponseError`. I think that the debug log should remain in `generateResponseError` as is. It will display the individual errors from the connector as well as the location in which they were aggregated.

bajtos: Fair enough :+1:

test/rest.test.js
```diff
@@ -1571,6 +1571,41 @@ describe('strong-remoting-rest', function() {
           .end(expectErrorResponseContaining({ message: 'an error' }, done));
       });
 
+      it('should return 500 if an array of errors is thrown', function(done) {
+        var testError = new Error('expected test error');
+        var errArray = [testError, testError];
+
+        var expectedMessage =
```
bajtos: I think this variable is no longer used, could you please remove?

test/rest.test.js
```diff
@@ -1571,6 +1571,41 @@ describe('strong-remoting-rest', function() {
...
+        request(app).get(method(testError).url)
+          .set('Accept', 'application/json')
+          .expect(500)
+          .end(function(err, res) {
+            if (err) return done(err);
+            expectedDetail = res.body.error;
+            delete expectedDetail.statusCode;
+
+            request(app).get(method(errArray).url)
+            .set('Accept', 'application/json')
```
bajtos: I think this line (and the following two lines should be indented by one level).

```js
request(app).get(...)
  .set(...)
  .expect(500)
  .end(...);
```

bajtos:
> I just thought there should be a more simple way to verify that the error is correct instead of nesting requests. 

I suppose you could simply check that the second request returns an error where `details` contains two entries and each entry has expected `message`? Such test will check less facts than your current implementation, but I think it should suffice.

Anyways, the current test looks good to me.

bajtos:
@richardpringle I found few nitpicks, see the comments above. The patch LGTM otherwise, no further review is necessary. Please back-port to 2.x too.

richardpringle:
@bajtos, changes made but I don't quite know what's going on with these failed tests...
There seem to be some unrelated error to my changes.

Otherwise this is ready to land.

richardpringle:
@slnode test please

richardpringle:
@slnode please test

richardpringle:
@bajtos, there are random dependency tests failing. I haven't looked at all of them but in connector-rest there are 3 tests failing only on node 0.10 on ubuntu and in lb-boot, linter is failing for node 5 on ubuntu and amazon. Linter is also failing on all platforms for lb-workspace.

I'll check to see if the master branch manages to pass CI but in the meantime, do you have any ideas?

richardpringle:
[This build](http://ci.strongloop.com/job/strong-remoting.pr/) hasn't passed in 6 months...

lib/rest-adapter.js
```diff
@@ -369,22 +369,40 @@ RestAdapter.errorHandler = function(options) {
...
-        data[prop] = err[prop];
+      if (Array.isArray(err)) {
+        var details = err.map(function(it) {
+          var data = generateResponseError(it);
+          delete data.statusCode;
+          return data;
+        });
+
+        err =
+          new Error('Failed with multiple errors, see `details` for more information.');
```
bajtos: Nitpick: I personally prefer this style:

```js
var err = new Error(
  'Failed with multiple errors, see `details` for more information.');
```

or even the following, which works better when the error message needs to be split onto multiple lines:

```js
var msg =  'Failed with multiple errors, see `details` for more information.';
var err = new Error(msg);
```

Feel free to ignore this comment.

bajtos:
LGTM.

As for the failing CI builds of dependencies, I hate to say it, but we should ignore those random failures and simply land this PR.

lib/rest-adapter.js
richardpringle: @bajtos, I will not ignore this comment! Haha, I completely agree. While I'm not a *huge* fan of your first suggestion, I do prefer your second. I usually do that, not sure why I didn't this time.

Regardless, I fixed it

PR #294, opened 2016-04-25
Set to no compression when using change stream

0candy:
Connect to https://github.com/strongloop/angular-live-set/issues/10


ritch:
LGTM 👍

PR #296, opened 2016-05-02
add strongErrorHandler configuration

loay:


loay:
@bajtos PTAL
related workspace PR: https://github.com/strongloop/loopback-workspace/pull/276

lib/jsonrpc-adapter.js
```diff
@@ -66,7 +66,7 @@ JsonRpcAdapter.prototype.getRoutes = function(obj) {
   return routes;
 };
 
-JsonRpcAdapter.errorHandler = function() {
+JsonRpcAdapter.strongErrorHandler = function() {
```
bajtos: What is the purpose of this API rename? Why can't we keep existing `errorHandler` name?

lib/rest-adapter.js
```diff
@@ -341,7 +341,7 @@ RestAdapter.urlNotFoundHandler = function() {
   };
 };
 
-RestAdapter.errorHandler = function(options) {
+RestAdapter.strongErrorHandler = function(options) {
```
bajtos: We need to change the implementation to use the new strong-error-handler module under the hood, the method names (API) can stay the same, no need to change that.

test/rest.test.js
```diff
@@ -45,7 +45,7 @@ describe('strong-remoting-rest', function() {
       process.env.NODE_ENV = 'test';
     }
     objects = RemoteObjects.create({ json: { limit: '1kb' },
-      errorHandler: { disableStackTrace: false }});
+      strongErrorHandler: { debug: true }});
```
bajtos: Ideally, we should preserve backwards compatibility in the sense that it is still possible to configure the error handler via `errorHandler: { disableStackTrace: false }}`.

Note that the fact that we use strong-error-handler under the hood is just an implementation detail that API consumers should not need to know about.

test/rest.test.js
```diff
@@ -110,7 +110,7 @@ describe('strong-remoting-rest', function() {
         }
       );
 
-      objects.options.errorHandler.handler = function(err, req, res, next) {
+      objects.options.strongErrorHandler.handler = function(err, req, res, next) {
```
bajtos: Ditto, this line should not need to be changed, the old API should continue to work in the new version.

test/rest.test.js
```diff
@@ -127,7 +127,7 @@ describe('strong-remoting-rest', function() {
     });
 
     it('should disable stack trace', function(done) {
-      objects.options.errorHandler.disableStackTrace = true;
+      objects.options.strongErrorHandler.debug = false;
```
bajtos: Ditto, this line should not need to be changed, the old API should continue to work in the new version.

If you prefer, then we can support both settings `debug:false` and `disableStackTrace:true`, but it's important to ensure that `disableStackTrace` keeps working.

However, while you are making this change, please change the default value of `disableStackTrace` to `true`, so that apps are configured for security by default.

test/rest.test.js
```diff
@@ -2227,7 +2227,6 @@ describe('strong-remoting-rest', function() {
         for (var prop in expected) {
           expect(result.body.error[prop], prop).to.equal(expected[prop]);
         }
-        expect(result.body.error.stack, 'stack').to.contain(__filename);
```
bajtos: Why is this change needed? I don't mind much, I just want to ensure we fully understand why we are making it.

bajtos:
@loay The idea is to rework strong-remoting to utilize the new strong-error-handler under the hood, so that we can get the benefit of higher security, while preserving public API as much as we can.

bajtos:
@loay @davidcheung PTAL please, I think this is no longer relevant, in which case we should close it.

davidcheung:
this should be covered by https://github.com/strongloop/strong-remoting/pull/302
going to close this PR 
cc @loay

Issue #299, opened 2016-05-12
type: 'any' in a remote method can coerce mongodb object ids into floats


vkarpov15:
If you have a remote method with a parameter that's `type: 'any'`, [coerce can convert a mongodb object id into a float](https://github.com/strongloop/strong-remoting/blob/aa391d2e65460196ae2a59493f9dd4844622f35d/lib/http-context.js#L296) because mongodb object id hex strings can happen to be all numbers. Easily fixed on my end (use `type: 'string'`) but this is an annoying rough edge in the API IMO.

richardpringle:
Hey @vkarpov15, those types are [LoopBack Types](https://docs.strongloop.com/display/public/LB/LoopBack+types); 'any' is not a type according to the documentation... And now I see that we test for `type: 'any'` <- @crandmck, think we could add that to the above link?

@vkarpov15, there are several issue with coercion that are currently being "cleaned up" by @bajtos , I will make sure that we consider this as well!

Cheers

richardpringle:
@bajtos, I keep adding the "coercion clean up" to related bugs. If you have a different workflow in mind please let me know. I'm not sure if your fixes will cover all cases, but at the very least I think we should add documentation for any remaining coercion behavior as a part of the milestone.

crandmck:
Yes, I can add that.  

Is the "Any" type name capitalized, as most of the other LoopBack types are?

Will it match just any primitive type, or will it also match Object, Array, and GeoPoint types?

vkarpov15:
It was lower case 'any' unfortunately - I assume the person that wrote it in the first place was anxious about whether the parameter would be a string or a full fledged mongodb ObjectId.

bajtos:
It's `any` and it basically means "anything" - a string, a number (used for IDs in SQL databases), or an object like ObjectID, or even an array or a boolean.

I am pondering an idea of always encoding the id property as strings, see https://github.com/strongloop/loopback/issues/2046

crandmck:
OK, I added it to the table in https://docs.strongloop.com/display/LB/LoopBack+types (and reordered the table alphabetically).

One final question: Will `any` match a `null` value?

vkarpov15:
IMO if you say "any" that should mean "no type coercion please, let me deal with it"

bajtos:
[...]

PR #301, opened 2016-05-16
allow disabling sketchy string coercion in config

vkarpov15:
Re: https://github.com/strongloop/loopback/issues/1217, https://github.com/strongloop/loopback/pull/1221, https://github.com/strongloop/strong-remoting/issues/299, etc. Ran into this bug yet again today. This coercion of string params is a major antipattern if you're using mongodb and there really needs to be a way to opt out of it.

### DIFF
--- a/test/http-context.test.js
+++ b/test/http-context.test.js
@@ -79,6 +79,12 @@ describe('HttpContext', function() {
         input: '000123',
         expectedValue: '000123',
       }));
+      it('skips coercion on "any" if user asks for it', givenMethodExpectArg({
+        type: 'any',
+        input: '200000000000000000000001',
+        expectedValue: '200000000000000000000001',
+        options: { coerce: false },
+      }));
     });
 
     describe('arguments with custom type', function() {
@@ -110,7 +116,7 @@ function givenMethodExpectArg(options) {
     var app = require('express')();
 
     app.get('/', function(req, res) {
-      var ctx = new HttpContext(req, res, method);
+      var ctx = new HttpContext(req, res, method, options.options);
       try {
         expect(ctx.args.testArg).to.eql(options.expectedValue);
       } catch (e) {
--- a/lib/http-context.js
+++ b/lib/http-context.js
@@ -201,7 +201,7 @@ HttpContext.prototype.buildArgs = function(method) {
     }
 
     // Coerce dynamic args when input is a string.
-    if (isAny && typeof val === 'string') {
+    if (isAny && typeof val === 'string' && this.options.coerce !== false) {
       val = coerceAll(val);
     }
 


slnode:

emgeee:
+1

diegonetto:
+1

alexhidalgo:
+1

richardpringle:
ok to test

richardpringle:
@vkarpov15, you need to sign the CLA for this to land!

vkarpov15:
@richardpringle where's the CLA?

richardpringle:
@superkhau, do you remember if the tests were passing CI on this PR merge: https://github.com/strongloop/loopback-connector-remote/commit/8343a649197f8a8388f35fcfc6f930aaae44d09f ? 

Same test keeps failing on difference builds.

Also, are there other tests that aren't running due to the failed dependent? I'm having the same issue with a PR on datasource-juggler.

superkhau:
Yeah, it was green when I merged it. However, when I check now, its partially red. I put in a fix here: https://github.com/strongloop/loopback-connector-remote/pull/47

stevegaossou:
+1

vkarpov15:
Anything else I can do to help get this merged? I'd really rather not have to write a separate access hook for every single model to work around this bug.

superkhau:
@richardpringle I believe remoting is all green again, just gotta wait for the LGTM from @bajtos at strongloop/loopback-connector-remote/pull/47

richardpringle:
@slnode test please

bajtos:
Hey @vkarpov15, thank you for the pull request, this is a neat solution. Just to let you know, our intention is to rework the coercion significantly to fix (hopefully all of) know issues. However, I think your solution is a great short-term improvement!

In order to get it landed, I need you to add a unit-test demonstrating correctness of your implementation and preventing regressions in the future.

vkarpov15:
@bajtos there's your test, this feature now has 100% coverage.

test/http-context.test.js
```diff
@@ -79,6 +79,12 @@ describe('HttpContext', function() {
         input: '000123',
         expectedValue: '000123',
       }));
+      it('skips coercion on "any" if user asks for it', givenMethodExpectArg({
```
bajtos: Please add an empty line before the test to serve as a visual delimiter.

bajtos:
Ah, I might have misunderstood the intention of this change originally, I though the option would be applied on per-argument basis.

If I understand it correctly, the new option is set globally. LoopBack users have two options how to enable the new feature:

1) In `server/config.json`, set:

```json
{
  "remoting": {
    "rest": {
      "coerce": false
    }
  }
}
```

2) In `server/middleware.json`:

```json
{
  "routes": {
     "loopback#rest": {
       "params": {
         "coerce": false
        }
     }
  }
}
```

@vkarpov15 is my understanding correct?

Considering that this new flag is global and controls only coercion of "any" types, I would like to use a different - more descriptive - option name, e.g. `coerceAnyType` or similar.

vkarpov15:
Yep your understanding is correct. I want a way to opt out of coercion of URL params at a global level. Feel free to make the changes you suggested on your end or close this PR - I've already put in way more time into this trivial one-liner than it's worth.

vkarpov15:
I'm gonna close this since I'm getting the feeling that this PR is not welcome. I'm just gonna hack around this for now with:

```javascript
const HttpContext = require('strong-remoting/lib/http-context')

const _httpContextBuildArgs = HttpContext.prototype.buildArgs
HttpContext.prototype.buildArgs = function (method) {
  const accepts = method.accepts
  accepts.forEach(param => {
    if (param && param.arg === 'id') {
      param.type = 'string'
    }
  })
  return _httpContextBuildArgs.call(this, method)
}
```

as a temporary fix while we work on the long-term solution of getting off of loopback.

PR #302, opened 2016-05-16
[SEMVER-MAJOR] Implement strong error handler for rest-adapter

davidcheung:
Using new strong-error-handler for rest-adapter

connect to https://github.com/strongloop-internal/scrum-loopback/issues/852

- [x] 3.0-RELEASE-NOTES.md

### DIFF
--- a/3.0-RELEASE-NOTES.md
+++ b/3.0-RELEASE-NOTES.md
@@ -27,3 +27,89 @@ To disable a static remote method
 
 To disable a prototype remote method:
 `disableMethodByName('prototype.updateAttributes')`
+
+## New error-handler for rest-adapter
+
+The REST adapter (typically created via `loopback.rest()`) uses a new error
+handler implementation that provides more secure configuration out of the box.
+
+The `error.status` has been removed and will be replaced by `error.statusCode`,
+`statusCode` is more descriptive and avoids ambiguity, users relying on `status`
+will need to change implementation accordingly.
+
+To replicate old behavior user can specify `config.local.js` as follow:
+```js
+module.exports = {
+  remoting : {
+    errorHandler: {
+      handler : function(err, req, res, defaultHandler) {
+        err.status = err.statusCode;
+        defaultHandler();
+      }
+    }
+  }
+}
+```
+
+The environment setting `NODE_ENV='production'` is no longer supported,
+production vs. debug mode is controlled exclusively by configuration.
+Production environment is assumed by default, the insecure debug mode
+must be explicitely turned on.
+
+You can learn more about the rationale behind the new handler in
+[this comment](https://github.com/strongloop/loopback/issues/1650#issuecomment-161920555)
+
+User can specific options in their applications in `config.json` as follow:
+```json
+{
+  "restApiRoot": "/api",
+  "host": "0.0.0.0",
+  "port": 3000,
+  "remoting": {
+    "errorHandler": {
+      "debug": false,
+      "log": false
+    }
+  }
+}
+```
+
+#### Production mode
+
+Stack trace is never returned in HTTP responses.
+
+Bad Request errors (4xx) provide the following properties copied from the
+error object: `name`, `message`, `statusCode` and `details`.
+
+All other errors (including non-Error values like strings or arrays) provide
+only basic information: `statusCode` and `message` set to status name from HTTP
+specification.
+
+#### Debug mode
+
+When in debug mode, HTTP responses include all properties provided by the error.
+
+For errors that are not an object, their string value is returned in
+`message` field.
+
+When a method fails with an array of errors (e.g. bulk create), HTTP the response
+contains still a single wrapper error with `details` set to the original array
+of errors.
+
+An example of an error response when an array of errors was raised:
+
+```js
+{
+  error: {
+    statusCode: 500,
+    name: 'ArrayOfErrors',
+    message: 'Failed with multiple errors, see `details` for more information.',
+    details: [
+      { name: 'Error1', message: 'expected error', statusCode: 500, stack: '<stacktrace>' },
+      { name: 'Error2', message: 'expected error2', statusCode: 500, stack: '<stacktrace>'}
+    ]
+  }
+}
+```
+
+Please see [Related code change](https://github.com/strongloop/strong-remoting/pull/302) here, and new [`strong-error-handler` here](https://github.com/strongloop/strong-error-handler/).
--- a/lib/rest-adapter.js
+++ b/lib/rest-adapter.js
@@ -29,6 +29,7 @@ var async = require('async');
 var HttpInvocation = require('./http-invocation');
 var ContextBase = require('./context-base');
 var HttpContext = require('./http-context');
+var strongErrorHandler = require('strong-error-handler');
 
 var json = bodyParser.json;
 var urlencoded = bodyParser.urlencoded;
@@ -333,7 +334,7 @@ RestAdapter.remoteMethodNotFoundHandler = function(className) {
     var message = 'Shared class "' + className + '"' +
       ' has no method handling ' + req.method + ' ' + req.url;
     var error = new Error(message);
-    error.status = error.statusCode = 404;
+    error.statusCode = 404;
     next(error);
   };
 };
@@ -342,13 +343,21 @@ RestAdapter.urlNotFoundHandler = function() {
   return function restUrlNotFound(req, res, next) {
     var message = 'There is no method to handle ' + req.method + ' ' + req.url;
     var error = new Error(message);
-    error.status = error.statusCode = 404;
+    error.statusCode = 404;
     next(error);
   };
 };
 
 RestAdapter.errorHandler = function(options) {
   options = options || {};
+  if (options.hasOwnProperty('disableStackTrace')) {
+    console.warn(
+      'strong-remoting no longer supports "errorHandler.disableStackTrace" option. ' +
+      'Use the new option "errorHandler.debug" instead.');
+  }
+
+  var strongErrorHandlerInstance = strongErrorHandler(options);
+
   return function restErrorHandler(err, req, res, next) {
     if (typeof options.handler === 'function') {
       try {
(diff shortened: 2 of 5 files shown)


test/rest.test.js
```diff
@@ -147,7 +149,7 @@ describe('strong-remoting-rest', function() {
         .end(expectErrorResponseContaining({ message: 'test-error' }, ['stack'], done));
     });
 
-    it('should disable stack trace', function(done) {
+    it.skip('should disable stack trace', function(done) {
```
davidcheung: this test is for `process.env.NODE_ENV = 'production';` 
do we still want to support this?

davidcheung:
@bajtos PTAL

bajtos:
@davidcheung I added two commits improving release notes, PTAL 751e708, 40e3d04. Feel free to squash them into your commit and dropping my authorship along the way.

3.0-RELEASE-NOTES.md
```diff
@@ -27,3 +27,67 @@ To disable a static remote method
...
+
+The environment setting `NODE_ENV='production'` is no longer supported,
+production vs. debug mode is controlled exclusively by configuration.
+Production environment is assumed by default, the insecure debug mode
+must be explicitely turned on.
+
+You can learn more about the rationale behind the new handler in
+[this comment](https://github.com/strongloop/loopback/issues/1650#issuecomment-161920555)
+
+User can specific options in their applications in `config.json` as follow:
```
bajtos: Are you sure about this? I think a better option is to configure strong-remoting's error handler via `server/middleware.json`, but we may not be there yet.

```json
{
  "routes": {
    "loopback#rest": {
      "params": {
        "debug": true,
        "log": true
      }
   }
}
```

3.0-RELEASE-NOTES.md
```diff
@@ -27,3 +27,67 @@ To disable a static remote method
...
+    statusCode: 500,
+    name: 'ArrayOfErrors',
+    message: 'Failed with multiple errors, see `details` for more information.',
+    details: [
+      { name: 'Error1', message: 'expected error', statusCode: 500, stack: '<stacktrace>' },
+      { name: 'Error2', message: 'expected error2', statusCode: 500, stack: '<stacktrace>'}
+    ]
+  }
+}
+```
```
bajtos: Please add a link to this pull request so that people can find the discussion that led to this change (see other release note entries) and also a pointer to strong-error-handler so that people can find more information about the settings available in this new error handler.

test/rest.test.js
bajtos: No, please remove this test.

test/rest.test.js
```diff
@@ -131,8 +133,9 @@ describe('strong-remoting-rest', function() {
         }));
     });
 
-    it('should disable stack trace', function(done) {
-      objects.options.errorHandler.disableStackTrace = true;
+    // skipped because new error handler does not have fine grain control for
+    // to retain message(debug mode) but have stacktrace off(!debug mode)
```
bajtos: We still need a test to verify that stack-trace is disabled by default. If it means the error message is different, then change the assert in the test to accommodate for this fact.

test/rest.test.js
```diff
@@ -55,6 +55,8 @@ describe('strong-remoting-rest', function() {
 
     // connect to the app
     objects.connect('http://localhost:' + server.address().port, adapterName);
+    var optsErrorHandler = { errorHandler: { debug: true, log: false }};
+    extend(objects.options, optsErrorHandler);
```
bajtos: Instead of extending `objects.options`, please modify the line 53 where the default errorHandler options are created.

This is what we have right now:

```js
objects = RemoteObjects.create({ json: { limit: '1kb' },
       errorHandler: { disableStackTrace: false }});
```

I believe we no longer support `disableStackTrace`, in which case it should be removed from tests too.

lib/rest-adapter.js
```diff
@@ -365,49 +366,7 @@ RestAdapter.errorHandler = function(options) {
...
-        }
-
-        data.stack = error.stack;
-        if (process.env.NODE_ENV === 'production' || options.disableStackTrace) {
-          delete data.stack;
-        }
-
-        return data;
-      }
+      return new strongErrorHandler(options)(err, req, res, next);
```
bajtos: This creates a new instance of `strongErrorHandler` for each failed request, which adds unnecessary performance overhead. Please cache the handler function created by `strongErrorHandler` factory.

bajtos: @davidcheung I would like you to take an extra step and warn users upgrading from 2.x to 3.0 when they forget to change the configuration of the error handler and keep using `option.disableStackTrace` (which no longer works).

My idea is modify `RestAdapter.errorHandler` and add a bit of code at the beginning that will detect unused option(s):

```js
RestAdapter.errorHandler = function(options) {
  options = options || {};
  if ('disableStackTrace' in options) {
    console.warn(
      'strong-remoting no longer supports "errorHandler.disableStackTrace" option. ' +
      'Use the new option "errorHandler.debug" instead.');
  }
  // etc.
};
```

bajtos:
@davidcheung reviewed.

3.0-RELEASE-NOTES.md
davidcheung: hmm it works in my sand-box
i think the behavior comes from https://github.com/strongloop/strong-remoting/blob/aa391d2e65460196ae2a59493f9dd4844622f35d/lib/rest-adapter.js#L315

davidcheung: maybe we can update this when we change it to use middleware.json?

test/rest.test.js
davidcheung: fixed

test/rest.test.js
davidcheung: added test

davidcheung:
@bajtos PTAL again

3.0-RELEASE-NOTES.md
bajtos: We are already using middleware.json. But never mind, let's leave this out of scope of this pull request and stay consistent with the current state of affairs (i.e. configure it via `server/config.json`).

3.0-RELEASE-NOTES.md
```diff
@@ -27,3 +27,69 @@ To disable a static remote method
...
+You can learn more about the rationale behind the new handler in
+[this comment](https://github.com/strongloop/loopback/issues/1650#issuecomment-161920555)
+
+User can specific options in their applications in `config.json` as follow:
+```json
+{
+  "restApiRoot": "/api",
+  "host": "0.0.0.0",
+  "port": 3000,
+  "errorHandler": {
```
bajtos: One more question though, I think `errorHandler` should be nested under `remoting`, could you please check?

```json
{
  "restApiRoot": "/api",
  "remoting": {
    "errorHandler": {
      "debug": true
    }
  }
}
```

lib/rest-adapter.js
```diff
@@ -346,8 +347,26 @@ RestAdapter.urlNotFoundHandler = function() {
   };
 };
 
+// singleton instance of strongErrorHandler
+RestAdapter.strongErrorHandler = null;
+
+// create or retrieve shared instance of strongErrorHandler
+RestAdapter.getStrongErrorHandler = function(options) {
```
bajtos: Sorry for the confusion. We should not use a true singleton here, because it's possible to create multiple `RestAdapter` instances, possibly configured with different `options`.

In my previous comments, I wanted you save the error handler instance in a local variable created inside `errorHandler` function but outside of `restErrorHandler` function.

Also I think `strongErrorHandler` was not meant to be called with `new`, since it returns another function, not an object.

```js
RestAdapter.errorHandler = function(options) {
  options = options || {};
  if (options.hasOwnProperty('disableStackTrace')) {
    // ...
  }

  var strongErrorHandlerInstance = strongErrorHandler(options);

  return function(restErrorHandler(err, req, res, next) {
    // ...
    strongErrorHandlerInstance(err, req, res, next);
  };
};
```

bajtos:
Looks mostly good, I left two more comments to address.

3.0-RELEASE-NOTES.md
davidcheung: you're right, the one I have was wrong, fixed.

davidcheung:
Thanks @bajtos! I have addressed your comments PTAL

3.0-RELEASE-NOTES.md
```diff
@@ -27,3 +27,71 @@ To disable a static remote method
...
+  "host": "0.0.0.0",
+  "port": 3000,
+  "remoting": {
+    "errorHandler": {
+      "debug": false,
+      "log": false
+    }
+  }
+}
+```
```
bajtos: Please add a link to strong-error-handler GH repository, so that people can easily find description of the config options available.

test/rest.test.js
```diff
@@ -131,36 +128,18 @@ describe('strong-remoting-rest', function() {
         }));
     });
 
-    it('should disable stack trace', function(done) {
-      objects.options.errorHandler.disableStackTrace = true;
+    it('should default with stack trace disabled', function(done) {
```
bajtos: `it('should exclude stack traces by default')`

3.0-RELEASE-NOTES.md
bajtos: Ah, I see the link is already included below, ignore this comment.

bajtos:
One nitpick, the rest LGTM. No further review is necessary.

@davidcheung this depends on https://github.com/strongloop/strong-error-handler/pull/11, doesn't it? Please wait with landing this patch until that required PR is landed first and CI builds are green on our Jenkins. (The jobs on Travis will keep failing because strong-error-handler is not published to npmjs.org yet.)

davidcheung:
yes what should i do with the `strong-error-handler` version on package.json?
>@davidcheung this depends on strongloop/strong-error-handler#11, doesn't it? Please wait with landing this patch until that required PR is landed first and CI builds are green on our Jenkins. (The jobs on Travis will keep failing because strong-error-handler is not published to npmjs.org yet.)

bajtos:
> yes what should i do with the strong-error-handler version on package.json?

Ah, I overlooked that line, please reset it to `^1.0.0`. I'll release `1.0.0` when your strong-error-handler pull request is landed.

davidcheung:
Please do not merge this yet, my loopback-core module is failing locally but the ci passed 
will probably need to submit another PR to loopback core before this should be merged.

davidcheung:
@slnode test please

davidcheung:
should land with https://github.com/strongloop/loopback/pull/2375 when ready

bajtos:
@davidcheung could you please add a note into release notes that `error.status` property is no longer available and clients should use `error.statusCode` instead?

davidcheung:
Placeholder: https://github.com/strongloop/loopback-datasource-juggler/blob/b039b51610b5ad14b5abdc5bb9e798baa7c70588/lib/dao.js#L1199
is still using `error.status`

bajtos:
LGTM

Amir-61:
@davidcheung Did this patch pass CI? Does this patch require [loopback-PR#2375](https://github.com/strongloop/loopback/pull/2375)? I The reason I'm asking is my PR in [loopback#2316](https://github.com/strongloop/loopback/pull/2316) was passing, but it fails with lots of failures now with [this commit](https://github.com/strongloop/strong-remoting/commit/5d0dcd2041e277ec1d81bc647c045e6518f43a63) in `strong-remoting'; I assume [loopback-PR#2375](https://github.com/strongloop/loopback/pull/2375) needs to be landed to fix these failures?

Thanks!

davidcheung:
@Amir-61 both PRs depend on each other to pass, I'm currently waiting for CI to run for [loopback-PR#2375](https://github.com/strongloop/loopback/pull/2375), once i merge that both sides should pass

PR #304, opened 2016-05-24
Test rest coercion

bajtos:
An integration test suite describing and verifying argument coercion in REST.
 - [x] arguments in query string
 - [x]  arguments in JSON body (formdata)
 - [x] arguments in url-encoded formdata
 - [x] the JSON body as an argument

```
1230 passing (3s)
```

(see https://github.com/strongloop-internal/scrum-loopback/issues/706#issuecomment-217371058 for more information)

In this pull request, I am describing the current behaviour in the `master` branch. My intention is to run the test suite against different strong-remoting versions to find differences, I'll post results later.

> In order to discover all breaking changes and document the new behaviour, we should write an extensive integration test suite that can be run against different versions of strong-remoting. Here is an initial draft for inspiration: https://gist.github.com/bajtos/b85e53a79ba06671e08d2013acccc611 Versions to run the suite against: v2.16.0, v2.16.1, latest 2.x, master (3.0-alpha). This commit may have introduced breaking change in v2.16.1: strongloop/strong-remoting@34b0665

Connect to strongloop-internal/scrum-loopback#884
Connect to strongloop-internal/scrum-loopback#913

@ritch @STRML please take a look

### DIFF
--- a/test/rest-coercion/README.md
+++ b/test/rest-coercion/README.md
@@ -1,0 +1,82 @@
+# Test coercion of input arguments in REST adapter
+
+The tests are grouped based on where the input argument is read from
+and what is the content-type of that source.
+
+After the tests finish, a CSV report is written in `report.csv`. This report
+makes it easy to compare results between different strong-remoting versions,
+just run `diff -du report1.csv report2.csv`.
+
+## Query string
+
+Arg spec: `{ arg: 'arg', http: { source: 'query' } }`
+
+Example request:
+
+```http
+GET /api?arg=value&arg2=value2 HTTP/1.1
+```
+
+See `urlencoded-*.suite.js` for tests.
+
+## URL-encoded form
+
+Arg spec: `{ arg: 'arg', http: { source: 'form' } }`
+
+Example request:
+
+```http
+POST /api HTTP/1.1
+Content-Type: application/x-www-form-urlencoded
+
+arg=value&arg2=value2
+```
+
+See `urlencoded-*.suite.js` for test cases.
+
+## JSON-encoded form
+
+Arg spec: `{ arg: 'arg', http: { source: 'form' } }`
+
+Example request:
+
+```http
+POST /api HTTP/1.1
+Content-Type: application/json
+
+{
+  "arg": "value",
+  "arg2": "value2"
+}
+```
+
+See `jsonform-*.suite.js` for test cases.
+
+## JSON-encoded body
+
+Arg spec: `{ arg: 'arg', http: { source: 'body' } }`
+
+Example request:
+
+```http
+POST /api HTTP/1.1
+Content-Type: application/json
+
+{
+  "name": "Superb",
+  "maker": "Skoda"
+}
+```
+
+See `jsonbody-*.suite.js` for test cases.
+
+The difference between JSON form and body is that the argument is set to full
+request body argument's http `source` is `body`.
+
+## URL-encoded body
+
+We don't have coverage for this scenario yet.
+
+## URL path parameters
+
+We don't have coverage for this scenario yet.
--- a/test/rest-coercion/_jsonbody.context.js
+++ b/test/rest-coercion/_jsonbody.context.js
@@ -1,0 +1,92 @@
+// Copyright IBM Corp. 2014,2016. All Rights Reserved.
+// Node module: strong-remoting
+// This file is licensed under the Artistic License 2.0.
+// License text available at https://opensource.org/licenses/Artistic-2.0
+
+'use strict';
+
+var debug = require('debug')('test');
+var expect = require('chai').expect;
+var util = require('util');
+var format = util.format;
+var extend = util._extend;
+
+module.exports = function createJsonBodyContext(ctx) {
+  return extend({
+    verifyTestCases: verifyTestCases,
+  }, ctx);
+
+  /**
+   * Verify a set of test-cases for a given argument specification
+   * (remoting definition).
+   *
+   * @param {Object} argSpec Argument definition, note that `http`
+   *   settings are injected automatically.
+   * @param {Array} testCases List of test cases to run & verify.
+   *   A test-case is a tuple [request-body, expected-argument-value]
+   *   The second item is optional, meaning the response should be the same
+   *   as the request.
+   *
+   * **Example**
+   *
+   * ```js
+   * verifyTestCases({ arg: 'name', type: 'any' }, [
+   *   [{ txt: null }],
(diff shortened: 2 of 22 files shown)


bajtos:
See also https://github.com/strongloop-internal/scrum-loopback/issues/706#issuecomment-217371058

STRML:
Looks good so far, I agree with the idea of testing this more thoroughly. And it does indeed appear 34b0665 broke some backcompat, but you could argue previous behavior was a bug in the first place. Still worthwhile for cleaning up for good in 3.0.

STRML:
See also https://github.com/strongloop/strong-remoting/pull/265 and especially https://github.com/strongloop/strong-remoting/pull/265#issuecomment-164527886 as handling this code was a comedy of errors - somebody inside strongloop force-pushed it out without any notice, then only half the PR was re-merged and released without review, causing significant breakage as the code was incomplete (it was two commits and should not have been merged incomplete).

bajtos:
Done. @STRML @ritch PTAL

test/rest-coercion/_jsonbody.context.js
```diff
@@ -0,0 +1,73 @@
+// Copyright IBM Corp. 2014,2016. All Rights Reserved.
```
ritch: whats with the `_` ?

test/rest-coercion/urlencoded-object.suite.js
```diff
@@ -0,0 +1,94 @@
...
+  suite('query string', urlEncodedContext(ctx, 'qs'));
+  suite('form data', urlEncodedContext(ctx, 'form'));
+};
+
+function suite(prefix, ctx) {
+  var EMPTY_QUERY = ctx.EMPTY_QUERY;
+  var ERROR_BAD_REQUEST = ctx.ERROR_BAD_REQUEST;
+  var verifyTestCases = ctx.verifyTestCases;
+
+  describe('query string - object - required', function() {
```
ritch: :clap: I love seeing this... (and the other suite with all these variable input tests)

test/helpers/builders.js
```diff
@@ -0,0 +1,24 @@
...
+'use strict';
+
+exports.remotingDate = function(val) {
+  return {
+    // NOTE(bajtos) The current implementation of strong-remoting
+    // calls `.toString()` which returns a date value like this:
+    //     Sat Jan 01 2000 00:00:00 GMT+0100 (CET)
+    // I think we should use `.toJSON()` instead, which returns concise
+    //     1999-12-31T23:00:00.000Z
+    '$data': new Date(val).toString(),
```
ritch: Yeah this should not use `toString()`. Is there a way to support both? Maybe a format setting somewhere. Not in this PR though.

test/rest-coercion/urlencoded-date.suite.js
```diff
@@ -0,0 +1,76 @@
...
+      ['arg=', INVALID_DATE], // should be: ERROR_BAD_REQUEST
+      ['arg=undefined', INVALID_DATE], // should be: ERROR_BAD_REQUEST
+      ['arg=null', INVALID_DATE], // should be: ERROR_BAD_REQUEST
+      ['arg=2016-05-19T13:28:51.299Z', remotingDate('2016-05-19T13:28:51.299Z')],
+    ]);
+  });
+
+  describe('query string - date - optional', function() {
+    verifyTestCases({ arg: 'arg', type: 'date' }, [
+      [EMPTY_QUERY, undefined],
```
ritch: Just a nit: its hard to tell what the eventual assertions are here... a comment would be enough.

test/rest-coercion/_jsonbody.context.js
bajtos: I have two sets of files: helper files that does not contain any tests (names starts with underscode), and ".suite.js" files with the tests. The code loading all tests files is skipping underscore files.

I don't mind using a different approach/convention, can you propose a better solution?

test/helpers/builders.js
bajtos: Let's discuss. I can see two approaches: 

1. A global feature flag that will control whether date is serialized via `toString()` or `toJSON()`. The application author decides the response format for all clients.

2. A special request argument (or HTTP header) allowing the client to request dates in the old strong-remoting@2.x format.

Thoughts?

test/rest-coercion/urlencoded-date.suite.js
bajtos: Done, see 20a40ce

test/helpers/builders.js
bajtos: I have reworked the tests to not depend on this any more. Let's move the discussion to https://github.com/strongloop/loopback/issues/1990.

bajtos:
I have cleaned up the code a lot, restructured test cases and added comments to make it more clear what is being tested. The patch is ready for another round of code review.

@STRML @ritch PTAL.

test/rest-coercion/README.md
```diff
@@ -0,0 +1,82 @@
...
+The difference between JSON form and body is that the argument is set to full
+request body argument's http `source` is `body`.
+
+## URL-encoded body
+
+We don't have coverage for this scenario yet.
+
+## URL path parameters
+
+We don't have coverage for this scenario yet.
```
bajtos: I created an internal story to keep track of the task of adding test coverage for URL path parameters, see https://github.com/strongloop-internal/scrum-loopback/issues/921

bajtos:
The high-level overview of changes between 2.16.x versions can be found here: https://gist.github.com/bajtos/8afaa9d1ed7d7b410c76ec2095a8a992

ritch:
LGTM

I don't see anything that should change. But there is quite a bit I may be overlooking still. I think the size of this test suite points to a separate module making sense for coercion.

The documentation in these tests would also be helpful to everyone. I think we should create a section about coercion in the readme that lists out the nuances of coercion. Which can be done outside of this PR.

Amir-61:
@slnode test please

Amir-61:
I could not find any problem either. The patch LGTM.

davidcheung:
LGTM too
should be really helpful for us to prevent regression on a much more detailed level :+1:

bajtos:
@ritch @Amir-61 @davidcheung thank you for the review!

> The documentation in these tests would also be helpful to everyone. I think we should create a section about coercion in the readme that lists out the nuances of coercion. Which can be done outside of this PR.

I filled a new task for that - https://github.com/strongloop/strong-remoting/issues/313, I think it will be best to document the _new_ behaviour after we make the intended changes/fixes.

PR #305, opened 2016-05-25
Implement getEndpoints

Amir-61:
This patch includes:

* Implement `getEndpoints` which returns an array of objects, showing verbs and paths
* Deprecate `getHttpMethod`
* Deprecate `getFullPath`

/to: @bajtos

### DIFF
--- a/lib/http-invocation.js
+++ b/lib/http-invocation.js
@@ -163,7 +163,8 @@ HttpInvocation.prototype._processArg = function(req, verb, query, accept) {
 
 HttpInvocation.prototype.createRequest = function() {
   var method = this.method;
-  var verb = method.getHttpMethod();
+  var endpoint = method.getEndpoints()[0];
+  var verb = endpoint.verb;
   var req = this.req = { method: verb || 'GET' };
   var accepts = method.accepts;
   var ctorAccepts = null;
@@ -171,7 +172,7 @@ HttpInvocation.prototype.createRequest = function() {
   var auth = this.auth;
 
   // initial url is the format
-  req.url = this.base + method.getFullPath();
+  req.url = this.base + endpoint.fullPath;
 
   var parsedUrl = urlUtil.parse(req.url);
 
--- a/lib/rest-adapter.js
+++ b/lib/rest-adapter.js
@@ -16,6 +16,7 @@ RestAdapter.RestMethod = RestMethod;
  * Module dependencies.
  */
 
+var deprecated = require('depd')('strong-remoting');
 var EventEmitter = require('events').EventEmitter;
 var debug = require('debug')('strong-remoting:rest-adapter');
 var util = require('util');
@@ -619,11 +620,30 @@ RestMethod.prototype.acceptsSingleBodyArgument = function() {
     getTypeString(accepts.type) == 'object' || false;
 };
 
+RestMethod.prototype.getEndpoints = function() {
+  var self = this;
+  return this.routes.map(function(route) {
+    var verbResult;
+    var verb = route.verb;
+    if (verb == 'all') {
+      verbResult = 'POST';
+    } else if (verb == 'del') {
+      verbResult = 'DELETE';
+    } else {
+      verbResult = verb.toUpperCase();
+    }
+    return {
+      verb: verbResult,
+      fullPath: joinPaths(self.restClass.getPath(), route.path),
+    };
+  });
+};
+
 RestMethod.prototype.getHttpMethod = function() {
-  var verb = this.routes[0].verb;
-  if (verb == 'all') return 'POST';
-  if (verb == 'del') return 'DELETE';
-  return verb.toUpperCase();
+  // deprecate message to let the users know what they were using
+  // was retuning just the first route's verb
+  deprecated('getHttpMethod() is deprecated, use getEndpoints()[0].verb instead.');
+  return this.getEndpoints()[0].verb;
 };
 
 RestMethod.prototype.getPath = function() {
@@ -631,7 +651,10 @@ RestMethod.prototype.getPath = function() {
 };
 
 RestMethod.prototype.getFullPath = function() {
-  return joinPaths(this.restClass.getPath(), this.getPath());
+  // deprecate message to let the users know what they were using
+  // was retuning just the first route's path
+  deprecated('getFullPath() is deprecated, use getEndpoints()[0].fullPath instead.');
+  return this.getEndpoints()[0].fullPath;
 };
 
 function getTypeString(ctorOrName) {
--- a/test/rest-adapter.test.js
+++ b/test/rest-adapter.test.js
@@ -13,6 +13,7 @@ var SharedClass = require('../lib/shared-class');
 var SharedMethod = require('../lib/shared-method');
 var expect = require('chai').expect;
 var factory = require('./helpers/shared-objects-factory.js');
+function NOOP() {};
 
 describe('RestAdapter', function() {
   var remotes;
@@ -218,6 +219,8 @@ describe('RestAdapter', function() {
     });
 
     describe('getHttpMethod', function() {
+      ignoreDeprecationsInThisBlock();
+
       it('returns POST for `all`', function() {
         var method = givenRestStaticMethod({ http: { verb: 'all' }});
         expect(method.getHttpMethod()).to.equal('POST');
@@ -245,6 +248,8 @@ describe('RestAdapter', function() {
     });
 
     describe('getFullPath', function() {
+      ignoreDeprecationsInThisBlock();
+
       it('returns class path + method path', function() {
         var method = givenRestStaticMethod(
           { http: { path: '/a-method' }},
@@ -252,6 +257,37 @@ describe('RestAdapter', function() {
         );
 
         expect(method.getFullPath()).to.equal('/a-class/a-method');
+      });
+    });
+
+    describe('getEndpoints', function() {
+      it('should return verb and fullPath for multiple paths', function() {
+        var method = givenRestStaticMethod({ http: [
+          { verb: 'DEL', path: '/testMethod1' },
+          { verb: 'PUT', path: '/testMethod2' },
+        ] });
+
+        var expectedEndpoints = [
+          {
+            fullPath: '/testClass/testMethod1',
+            verb: 'DELETE',
+          }, {
+            fullPath: '/testClass/testMethod2',
+            verb: 'PUT',
+          },
+        ];
+
+        expect(method.getEndpoints()).to.eql(expectedEndpoints);
+      });
+
+      it('should return verb and fullPath for single path', function() {
+        var method = givenRestStaticMethod({ http: { verb: 'all' }});
+        expect(method.getEndpoints()).to.eql([
+          {
+            verb: 'POST',
+            fullPath: '/testClass/testMethod',
+          },
+        ]);
       });
     });
 
@@ -438,4 +474,14 @@ describe('RestAdapter', function() {
 });
 
 function someFunc() {
+}
+
+function ignoreDeprecationsInThisBlock() {
+  before(function() {
+    process.on('deprecation', NOOP);
+  });
+
+  after(function() {
+    process.removeListener('deprecation', NOOP);
+  });
 }


bajtos:
I am proposing to preserve backwards compatibility of this API to make migration easier, and provide new methods for the functionality we need. We can mark the old methods as deprecated, it will make it easier to find all places calling them.

```js
RestMethod.prototype.getHttpMethod = function() {
  deprecated('getHttpMethod() is deprecated, use getFirstHttpMethod() instead.');
  return this.getFirstHttpMethod();
};

RestMethod.prototype.getFirstHttpMethod = function() {
  return this.getAllHttpMethods()[0];
};

RestMethod.prototype.getAllHttpMethods = function() {
  // the real works is done here
};
```

test/rest-adapter.test.js
```diff
@@ -220,17 +220,17 @@ describe('RestAdapter', function() {
     describe('getHttpMethod', function() {
       it('returns POST for `all`', function() {
         var method = givenRestStaticMethod({ http: { verb: 'all' }});
-        expect(method.getHttpMethod()).to.equal('POST');
+        expect(method.getHttpMethod()[0]).to.equal('POST');
```
bajtos: With the ability to get all methods in place, we should test all members of the array returned.

```js
expect(method.getAllHttpMethods()).to.eql(['POST']);
```

Amir-61:
@slnode test please

Amir-61:
@slnode test please

Amir-61:
Hey @rmg 

I see lots of failures for dependencies like the following. Any thought about this? Thanks!

```
npm ERR! Error: connect ECONNREFUSED 10.0.1.45:32770
npm ERR!     at Object.exports._errnoException (util.js:893:11)
npm ERR!     at exports._exceptionWithHostPort (util.js:916:20)
npm ERR!     at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1075:14)
npm ERR!  { [Error: connect ECONNREFUSED 10.0.1.45:32770]
npm ERR!   code: 'ECONNREFUSED',
npm ERR!   errno: 'ECONNREFUSED',
npm ERR!   syscall: 'connect',
npm ERR!   address: '10.0.1.45',
npm ERR!   port: 32770,
npm ERR!   parent: 'debug' }
npm ERR! 
npm ERR! If you are behind a proxy, please make sure that the
npm ERR! 'proxy' config is set properly.  See: 'npm help config'
```

Amir-61:
@slnode test please

rmg:
When that temporary registry is killed, anything trying to install from it will fail. This can happen if one of the dependents fails and triggers a registry shutdown.

Amir-61:
@slnode test please

Amir-61:
@slnode test please

Amir-61:
@bajtos PTAL :-)

lib/rest-adapter.js
```diff
@@ -619,19 +620,49 @@ RestMethod.prototype.acceptsSingleBodyArgument = function() {
...
-  if (verb == 'all') return 'POST';
-  if (verb == 'del') return 'DELETE';
-  return verb.toUpperCase();
+  // depricate message to let the users know what they were using
+  // was retuning just the first route's verb
+  deprecated('getHttpMethod() is deprecated, use getFirstHttpMethod() instead.');
+  return this.getFirstHttpMethod();
+};
+
+RestMethod.prototype.getFirstHttpMethod = function() {
```
bajtos: On the second thought, I think `getFirstHttpMethod` may be not worth the API pollution. Users can always call `this.getEndpoints()[0].verb` themselves. What do you think?

lib/rest-adapter.js
```diff
@@ -619,19 +620,49 @@ RestMethod.prototype.acceptsSingleBodyArgument = function() {
...
+    };
+  });
+};
+
 RestMethod.prototype.getHttpMethod = function() {
-  var verb = this.routes[0].verb;
-  if (verb == 'all') return 'POST';
-  if (verb == 'del') return 'DELETE';
-  return verb.toUpperCase();
+  // depricate message to let the users know what they were using
```
bajtos: typo: `depricate`, should be `deprecate`.

lib/rest-adapter.js
```diff
@@ -619,19 +620,49 @@ RestMethod.prototype.acceptsSingleBodyArgument = function() {
...
 };
 
 RestMethod.prototype.getPath = function() {
   return this.routes[0].path;
 };
 
 RestMethod.prototype.getFullPath = function() {
-  return joinPaths(this.restClass.getPath(), this.getPath());
+  // depricate message to let the users know what they were using
+  // was retuning just the first route's path
```
bajtos: ditto

lib/rest-adapter.js
```diff
@@ -619,19 +620,49 @@ RestMethod.prototype.acceptsSingleBodyArgument = function() {
...
+  return this.getEndpoints()[0].verb;
 };
 
 RestMethod.prototype.getPath = function() {
   return this.routes[0].path;
 };
 
 RestMethod.prototype.getFullPath = function() {
-  return joinPaths(this.restClass.getPath(), this.getPath());
+  // depricate message to let the users know what they were using
```
bajtos: ditto, the same typo here

bajtos:
@Amir-61 please add a test to verify that `getEndpoints` returns all paths.

Also what happens when you run the test suite, does it trigger deprecation warnings because we still call `getMethod`?

rmg:
@slnode test please

Amir-61:
@bajtos Please review another round :-)

>Also what happens when you run the test suite, does it trigger deprecation warnings because we still call `getMethod`?

If you mean `getHttpMethod` and `getFullPath`, yes they show the deprecated messages. I personally think it is good to show deprecated message to let the users know what they were using
 was retuning just the first route's verb/path. What is your thought?

Also, this [generateRestMethodWithMultiplePaths](https://github.com/strongloop/strong-remoting/pull/305/files#diff-cbc4ab1c0c4764b3a7b2a66ebc51bdabR286) may be messy. would you do it in the same way? How would you do that?

Amir-61:
@slnode test please

bajtos:
> If you mean getHttpMethod and getFullPath, yes they show the deprecated messages. I personally think it is good to show deprecated message to let the users know what they were using
was retuning just the first route's verb/path. What is your thought?

It's good to show deprecation messages for strong-remoting users, but it's not great to see them when running a test suite.

I am proposing the following:
 - Rework all tests using deprecated methods to use the new official methods
 - In unit-tests that are testing specifically these deprecated methods, temporarily disable deprecation messages via https://github.com/dougwilson/nodejs-depd#processenvno_deprecation

> Also, this generateRestMethodWithMultiplePaths may be messy. would you do it in the same way? How would you do that?

 I would simply call the existing helper `givenRestStaticMethod` instead of writing a new helper:

```js
givenRestStaticMethod({ http: [
  { verb: 'DEL', path: '/testMethod1' },
  { verb: 'PUT', path: '/testMethod2' }
]});
```

Unless I am missing a reason why this would not work?

BTW I think we should backport this patch to 2.x too.

test/rest-adapter.test.js
```diff
@@ -217,20 +217,20 @@ describe('RestAdapter', function() {
...
         var method = givenRestStaticMethod({ http: { verb: 'del' }});
-        expect(method.getHttpMethod()).to.equal('DELETE');
+        expect(method.getEndpoints()[0].verb).to.equal('DELETE');
       });
 
       it('returns upper-case value otherwise', function() {
         var method = givenRestStaticMethod({ http: { verb: 'get' }});
-        expect(method.getHttpMethod()).to.equal('GET');
+        expect(method.getEndpoints()[0].verb).to.equal('GET');
       });
```
Amir-61: >I am proposing the following:

>* Rework all tests using deprecated methods to use the new official methods
* In unit-tests that are testing specifically these deprecated methods, temporarily disable deprecation messages via https://github.com/dougwilson/nodejs-depd#processenvno_deprecation

Here I tried to use the new official methods in tests; for instance using `getEndpoints()[0].verb` instead of `getHttpMethod()` which is now deprecated.

However I believe what I did is not what you proposed. IIUC, here in this test that we are explicitly testing this deprecated method, I should just have them untouched, but without deprecation message. However, based on [other usages](https://github.com/strongloop/loopback/blob/6964914bab64496c789c434fec39d2914231ee23/Gruntfile.js#L9) I tried `process.env.NO_DEPRECATION = 'strong-remoting';` but still it shows the deprecation messages...

test/rest-adapter.test.js
```diff
@@ -17,6 +17,10 @@ var factory = require('./helpers/shared-objects-factory.js');
 describe('RestAdapter', function() {
   var remotes;
 
+  before(function() {
+    process.env.NO_DEPRECATION = 'strong-remoting';
+  });
+
```
Amir-61: >I am proposing the following:

>* Rework all tests using deprecated methods to use the new official methods
* In unit-tests that are testing specifically these deprecated methods, temporarily disable deprecation messages via https://github.com/dougwilson/nodejs-depd#processenvno_deprecation

I reworked on [where](https://github.com/strongloop/strong-remoting/pull/305/files#diff-bc246aed6f72bde8d7cc584ef435426aR166) we were using deprecated `getHttpMethod` and `getPath` to use our new official methods, but [here](https://github.com/strongloop/strong-remoting/blob/ac3093dcfbb787977ca0229b0f672703859e52e1/test/rest-adapter.test.js#L220-L245) that we we are explicitly testing these deprecated methods, I should just have them untouched, and need to disable deprecation messages temporarily. However, I tried `process.env.NO_DEPRECATION = 'strong-remoting';` but still it shows the deprecation messages...

Also it is worth mentioning that there are many other deprecated messages for other methods (not `getHttpMethod` or `getPath`)which are shown when you run tests for `strong-remoting`, I did not touch them because I thought it is out of scope of this PR.

Amir-61:
@slnode test please

test/rest-adapter.test.js
bajtos: Oh, I think I know what's the problem: `depd` initializes when the file is `require`-d, thus changing `env.NO_DEPRECATION` at runtime has no effect. Can you try to use `process.on('deprecation')` instead? See [docs](https://www.npmjs.com/package/depd#processondeprecation-fn).

lib/http-invocation.js
```diff
@@ -163,15 +163,15 @@ HttpInvocation.prototype._processArg = function(req, verb, query, accept) {
 
 HttpInvocation.prototype.createRequest = function() {
   var method = this.method;
-  var verb = method.getHttpMethod();
+  var verb = method.getEndpoints()[0].verb;
```
bajtos: perhaps store the endpoint, so that we don't have to repeat `method.getEndpoints[0]`?

```js
var endpoint = method.getEndpoints()[0]);
var verb = endpoints.verb;
//...
req.url = this.base  + endpoint.fullPath;
```

Thoughts?

test/rest-adapter.test.js
bajtos: You also need an `after` handler to reset the environment (remove the listener), so that deprecations are printed again.

test/rest-adapter.test.js
```diff
@@ -18,7 +18,15 @@ describe('RestAdapter', function() {
   var remotes;
 
   before(function() {
-    process.env.NO_DEPRECATION = 'strong-remoting';
+    process.on('deprecation', function() {
+      process.env.NO_DEPRECATION = 'strong-remoting';
```
bajtos: No need to modify ENV, the deprecation listener should be a no-op here.

test/rest-adapter.test.js
```diff
@@ -18,7 +18,15 @@ describe('RestAdapter', function() {
...
+    });
+  });
+
+  after(function() {
+    process.removeListener('deprecation', function() {
```
bajtos: I think this will create a different function instance than the one that you passed in the `before` hook. This is the correct solution IMO:

```
before(function() {
  process.on('deprecation', NOOP);
});

after(function() {
  process.removeListener('deprecation', NOOP);
});

function NOOP() {}
```

bajtos:
@Amir-61 LGTM.

The current version disables deprecation warnings printed from all tests in `test/rest-adapter.test.js`, which may hide deprecations we actually want to be aware of. A better solution is to disable deprecation warnings only for those few tests where we are intentionally calling deprecated methods directly from the test. Let's do that in a new PR though, so that you can work on other things depending on the new API added by this PR too.

Also I think the new API should be back-ported to 2.x too, possibly without deprecation warnings. Thoughts?

Amir-61:
@slnode test please

Amir-61:
@bajtos I did not want to leave you unhappy so I did not land the PR to make sure you are happy with :-) I fixed it now please take a look again.

test/rest-adapter.test.js
```diff
@@ -245,6 +254,13 @@ describe('RestAdapter', function() {
     });
 
     describe('getFullPath', function() {
+      before(function() {
+        process.on('deprecation', NOOP);
+      });
+
+      after(function() {
+        process.removeListener('deprecation', NOOP);
+      });
```
bajtos: Nitpick: add an empty line to visually delimit after/it blocks.

Idea to consider: extract these two before/after blocks into a shared helper method.

```js
describe('getFullPath', function() {
  ignoreDeprecationsInThisBlock();

  it('returns class path + method path', /*...*/);
});
```

Thoughts?

test/rest-adapter.test.js
Amir-61: I addressed this and pushed the change. Thanks!

test/rest-adapter.test.js
Amir-61: I addressed this and pushed the change. Thanks!

Amir-61:
@slnode test please

bajtos:
Lovely 👏 

Could you please fix commit message summary to use `getEndpoints` instead of `getEndPoints`, in order to match the real API? Also the message body has typo "Deprectae" - should be "Deprecate".

Last but not least, please check why some of the dependants are failing, to make sure the failures are not related.

No further review is necessary.

Amir-61:
@bajtos 
>Last but not least, please check why some of the dependants are failing, to make sure the failures are not related.

Unfortunately, failures are so random and they are not code relevant either. They all fail with this error: `[sl-ci-run ERROR 70.64 ]: Linter failed.`. Please see [this](http://ci.strongloop.com/job/loopback/node=0.12,os=amazon/3925/console), [this](http://ci.strongloop.com/job/loopback-connector-remote/node=4.x,os=ubuntu/1357/console), [this](http://ci.strongloop.com/job/strongloop/node=0.12,os=amazon/5865/console),... The same code on weekend passed about 47/50 tests, but now they are not passing; I guess the busier Jenkis is, the lower chance to pass tests :(

If I run many times by chance one of them may pass. I add the ci failures in our `CI Failure report` GH issue

Amir-61:
@slnode test please

Issue #311, opened 2016-06-07
Prototype methods apply ACL checks only after a model instance is found.


richardpringle:
When making a call to `VERB api/ExampleModel/:id/operation` the ACL is only checked upon model instance discovery, otherwise you get a 404 response instead of a 401. This indicates that a record with that particular `id` does not exist even if the user does not have permission to check whether particular models exists or not.

bajtos:
Cross posting @michaelfreund's  https://github.com/strongloop/loopback/issues/2398#issuecomment-224303992

> it is an information leak, where model ids can be iterated to check for model existence. this might be necessary information for further attack.

bajtos:
@michaelfreund  @zkuang @richardpringle I have submitted to alternative pull requests to fix this issue, see #315 and #316. We are discussing with @ritch what would be the best fix, our concern is backwards compatibility. Feel free to try those patches out in your app and/or leave a comment what do you think.

bajtos:
The fix is available in `strong-remoting@2.28.1`.

michaelfreund:
@bajtos we can confirm this fixes #2220 for us.

zkuang:
@bajtos sorry for the late reply. I'll check them out. Thanks very much.

PR #315, opened 2016-06-20
Move sharedCtor call after auth checks

bajtos:
Possible breaking change (only prototype methods are affected):

Any code run before "invoke:before" phase (where we initialise the
instance) will not receive `ctx.instance`:

 - "remotes.authorization" hook
 - "rest.before" hooks
 - custom phases inserted before "invoke" phase

Connect to #311

@ritch please review ASAP.


bajtos:
Closing in favour of #316

PR #316, opened 2016-06-20
Prioritise auth errors over sharedCtor errors

bajtos:
Return 401 Not Authorized when making an unauthorised request to a prototype method for a model instance that does not exist.

Potentially breaking change affecting prototype methods:

Hooks registered before "invoke" phase (i.e. "auth" phase and any custom
phases inserted before "invoke") will be called even when the model
instance was not found. The property "ctx.instance" will be undefined
in such case, hook handlers must account for this case.

This is an alternative to #315 that may have less breaking changes.

Connect to #311

@ritch please review.

### DIFF
--- a/lib/remote-objects.js
+++ b/lib/remote-objects.js
@@ -642,6 +642,10 @@ RemoteObjects.prototype._setupPhases = function() {
     self._executeAuthorizationHook(ctx, next);
   });
 
+  invoke.before(function reportSharedCtorError(ctx, next) {
+    next(ctx.sharedCtorError);
+  });
+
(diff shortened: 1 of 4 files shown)


bajtos:
Closing in favour of #316

ritch:
LGTM

ritch:
> Closing in favour of #316

??? I think you meant to close #315

bajtos:
>  I think you meant to close #315

Ah, of course, I have reopened this one.

bajtos:
Landed and released in `strong-remoting@2.28.1`

PR #323, opened 2016-07-20
Introduce support for integer datatype

### LINKED ISSUE #317: Support for `integer` datatype removed from loopback\strong-remoting
IBM support sent me this way. I had created IBM service request 46011999000 because this is for GoDaddy and we have a paid subscription.

loopback version 2.18.1
strong-remoting version 2.19.0
 
converting parameter to target type includes explicit support for 'integer' in shared-method.js
lines 390-443

```node 
function convertToBasicRemotingType(type) {
  if (Array.isArray(type)) {
    return type.map(convertToBasicRemotingType);
  }
 
  if (typeof type === 'object') {
    type = type.modelName || type.name;
  }
 
  type = String(type).toLowerCase();
 
  switch (type) {
    case 'string':
    case 'number':
    case 'date':
    case 'boolean':
    case 'buffer':
    case 'object':
    case 'any':
      return type;
    case 'integer':
      return 'number';
    case 'array':
      return ['any'].map(convertToBasicRemotingType);
    default:
      // custom types like MyModel
      return 'object';
  }
}
 
function convertValueToTargetType(argName, value, targetType) {
  switch (targetType) {
    case 'string':
      return String(value).valueOf();
    case 'date':
      return new Date(value);
    case 'number':
    case 'integer':
      return Number(value).valueOf();
    case 'boolean':
      return Boolean(value).valueOf();
    // Other types such as 'object', 'array',
    // ModelClass, ['string'], or [ModelClass]
    default:
      switch (typeof value) {
        case 'string':
          return JSON.parse(value);
        case 'object':
          return value;
        default:
          throw new badArgumentError(argName + ' must be ' + targetType);
      }
  }
}
```
 
loopback version 2.28.0
strong-remoting version 2.26.0
 
converting parameter to target type does not include support for 'integer' in shared-method.js
lines 373-424
 
```node
function convertToBasicRemotingType(type) {
  if (Array.isArray(type)) {
    return type.map(convertToBasicRemotingType);
  }
 
  if (typeof type === 'object') {
    type = type.modelName || type.name;
  }
 
  type = String(type).toLowerCase();
 
  switch (type) {
    case 'string':
    case 'number':
    case 'date':
    case 'boolean':
    case 'buffer':
    case 'object':
    case 'file':
    case 'any':
      return type;
    case 'array':
      return ['any'].map(convertToBasicRemotingType);
    default:
      // custom types like MyModel
      return 'object';
  }
}
 
function convertValueToTargetType(argName, value, targetType) {
  switch (targetType) {
    case 'string':
      return String(value).valueOf();
    case 'date':
      return new Date(value);
    case 'number':
      return Number(value).valueOf();
    case 'boolean':
      return Boolean(value).valueOf();
    // Other types such as 'object', 'array',
    // ModelClass, ['string'], or [ModelClass]
    default:
      switch (typeof value) {
        case 'string':
          return JSON.parse(value);
        case 'object':
          return value;
        default:
          throw new badArgumentError(argName + ' must be ' + targetType);
      }
  }
}
```

We have Swagger specs that have been using integer which is completely valid but a newer version of loopback we'd be forced to switch to number and manually handle checking for fractional values. I don't think that's appropriate.

gunjpan:
all types of numbers, in Javascript, are of only one dataType: `Numbers`.
~~This is a POC~~ Patch to fix #317.
~~Note: Breaks few test cases on local~~

Connect to #317

### DIFF
--- a/lib/shared-method.js
+++ b/lib/shared-method.js
@@ -285,6 +285,12 @@ function badArgumentError(msg) {
   return err;
 }
 
+function internalServerError(msg) {
+  var err = new Error(msg);
+  err.statusCode = 500;
+  return err;
+}
+
 function escapeRegex(d) {
   // see http://stackoverflow.com/a/6969486/69868
   return d.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, '\\$&');
@@ -318,7 +324,7 @@ function coerceAccepts(uarg, desc) {
     });
   }
 
-  var actualType = SharedMethod.getType(uarg);
+  var actualType = SharedMethod.getType(uarg, targetType);
 
   // convert values to the correct type
   // TODO(bajtos) Move conversions to HttpContext (and friends)
@@ -331,7 +337,7 @@ function coerceAccepts(uarg, desc) {
     // JSON.parse can throw, so catch this error.
     try {
       uarg = convertValueToTargetType(name, uarg, targetType);
-      actualType = SharedMethod.getType(uarg);
+      actualType = SharedMethod.getType(uarg, targetType);
     } catch (e) {
       var message = util.format('invalid value for argument \'%s\' of type ' +
         '\'%s\': %s. Received type was %s. Error: %s',
@@ -369,7 +375,14 @@ function coerceAccepts(uarg, desc) {
   if (actualType === 'number' && Number.isNaN(uarg)) {
     throw new badArgumentError(name + ' must be a number');
   }
-
+  if (actualType === 'integer') {
+    if (Number.isNaN(uarg)) {
+      throw new badArgumentError(name + ' must be an integer');
+    }
+    if (!Number.isSafeInteger(uarg)) {
+      throw new badArgumentError(name + ' must be a safe integer');
+    }
+  }
   return uarg;
 }
 
@@ -395,6 +408,7 @@ function convertToBasicRemotingType(type) {
   switch (type) {
     case 'string':
     case 'number':
+    case 'integer':
     case 'date':
     case 'boolean':
     case 'buffer':
@@ -417,6 +431,7 @@ function convertValueToTargetType(argName, value, targetType) {
     case 'date':
       return new Date(value);
     case 'number':
+    case 'integer':
       return Number(value).valueOf();
     case 'boolean':
       return Boolean(value).valueOf();
@@ -440,16 +455,18 @@ function convertValueToTargetType(argName, value, targetType) {
  * @returns {String} The type name
  */
 
-SharedMethod.getType = function(val) {
+SharedMethod.getType = function(val, targetType) {
   var type = typeof val;
 
   switch (type) {
     case 'undefined':
     case 'boolean':
-    case 'number':
     case 'function':
     case 'string':
       return type;
+    case 'number':
+      return Number.isInteger(val) && targetType === 'integer' ?
+        'integer' : 'number';
     case 'object':
       // null
       if (val === null) {
@@ -541,8 +558,9 @@ SharedMethod.toResult = function(returns, raw, ctx) {
     }
 
     if (item.root) {
-      var isFile = convertToBasicRemotingType(item.type) === 'file';
-      result = isFile ? raw[index] : convert(raw[index]);
+      var targetType = convertToBasicRemotingType(item.type);
+      var isFile =  targetType === 'file';
+      result = isFile ? raw[index] : convert(raw[index], targetType, item.name);
       return false;
     }
 
@@ -551,21 +569,34 @@ SharedMethod.toResult = function(returns, raw, ctx) {
 
   returns.forEach(function(item, index) {
     var name = item.name || item.arg;
-    if (convertToBasicRemotingType(item.type) === 'file') {
+    var targetType = convertToBasicRemotingType(item.type);
+    if (targetType === 'file') {
       console.warn('%s: discarded non-root return argument %s of type "file"',
         this.stringName,
         name);
       return;
     }
 
-    var value = convert(raw[index]);
+    var value = convert(raw[index], targetType, name);
     result[name] = value;
   });
 
   return result;
 
-  function convert(val) {
-    switch (SharedMethod.getType(val)) {
+  function convert(val, targetType, argName) {
+    var targetTypeIsIntegerArray = Array.isArray(targetType) &&
+      targetType.length === 1 && targetType[0] === 'integer';
+
+    if (targetTypeIsIntegerArray && Array.isArray(val)) {
+      var arrayTargetType = targetType[0];
+      return val.map(function(intVal) {
+        return convert(intVal, arrayTargetType, argName);
+      });
+    }
+
+    var actualType = SharedMethod.getType(val, targetType);
+
+    switch (actualType) {
       case 'date':
         return {
           $type: 'date',
@@ -576,9 +607,29 @@ SharedMethod.toResult = function(returns, raw, ctx) {
           $type: 'base64',
           $data: val.toString('base64'),
         };
-    }
+      default:
+        if (targetType === 'integer') {
+          var message;
 
-    return val;
+          if (!Number.isInteger(val)) {
+            message = util.format(
+              'Invalid return value for argument \'%s\' of type ' +
+              '\'%s\': %s. Received type was %s.',
+              argName, targetType, val, typeof val);
+            throw internalServerError(message);
+          }
+
+          if (!Number.isSafeInteger(val)) {
+            message = util.format(
+              'Unsafe integer value returned for argument \'%s\' of type ' +
+              '\'%s\': %s.',
+              argName, targetType, val);
+            throw internalServerError(message);
+          }
+          return val;
+        }
+        return val;
+    }
   }
 };
 
@@ -664,4 +715,17 @@ SharedMethod.prototype.addAlias = function(alias) {
   if (this.aliases.indexOf(alias) === -1) {
     this.aliases.push(alias);
   }
+};
+
+// To support Node v.0.10.x
+Number.isInteger = Number.isInteger || function(value) {
+  return typeof value === 'number' &&
+    isFinite(value) &&
+    Math.floor(value) === value;
+};
+
+Number.MAX_SAFE_INTEGER = Number.MAX_SAFE_INTEGER || Math.pow(2, 53) - 1;
+
+Number.isSafeInteger = Number.isSafeInteger || function(value) {
+  return Number.isInteger(value) && Math.abs(value) <= Number.MAX_SAFE_INTEGER;
 };
--- a/test/rest-coercion/jsonform-integer.suite.js
+++ b/test/rest-coercion/jsonform-integer.suite.js
@@ -1,0 +1,79 @@
+// Copyright IBM Corp. 2014,2016. All Rights Reserved.
+// Node module: strong-remoting
+// This file is licensed under the Artistic License 2.0.
+// License text available at https://opensource.org/licenses/Artistic-2.0
+
+'use strict';
+
+var jsonFormContext = require('./_jsonform.context');
+
+module.exports = function(ctx) {
+  ctx = jsonFormContext(ctx);
+  var EMPTY_BODY = ctx.EMPTY_BODY;
+  var ERROR_BAD_REQUEST = ctx.ERROR_BAD_REQUEST;
+  var verifyTestCases = ctx.verifyTestCases;
+
+  describe('json form - integer - required', function() {
+    // See verifyTestCases' jsdoc for details about the format of test cases.
+    verifyTestCases({ arg: 'arg', type: 'integer', required: true }, [
+      // Valid values
+      [{ arg: 0 }, 0],
+      [{ arg: 1 }, 1],
+      [{ arg: -1 }, -1],
+
+      // Empty values should trigger ERROR_BAD_REQUEST
+      [EMPTY_BODY, ERROR_BAD_REQUEST],
+      [{ arg: null }, 0], // should be: ERROR_BAD_REQUEST
+      [{ arg: '' }, 0], // should be: ERROR_BAD_REQUEST
+    ]);
+  });
+
+  describe('json form - integer - optional', function() {
+    // See verifyTestCases' jsdoc for details about the format of test cases.
+    verifyTestCases({ arg: 'arg', type: 'integer' }, [
+      // Empty values
+      [EMPTY_BODY, undefined],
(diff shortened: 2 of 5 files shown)


gunjpan:
@bajtos : Here is the PoC for introducing `integer` data types. PTAL and let's discuss if the direction is correct. Please note, CI will fail as it had six failing test cases on local, which will be fixed if we're moving further this direction

lib/shared-method.js
```diff
@@ -440,16 +444,17 @@ function convertValueToTargetType(argName, value, targetType) {
...
 
   switch (type) {
     case 'undefined':
     case 'boolean':
-    case 'number':
     case 'function':
     case 'string':
       return type;
+    case 'number':
+      return Number.isInteger(val) ? (targetType || 'integer') : 'number';
```
gunjpan: that's the hacky way to work with numbers which are of pattern: `x.0`, for example: `2.0`

lib/shared-method.js
```diff
@@ -369,7 +369,9 @@ function coerceAccepts(uarg, desc) {
   if (actualType === 'number' && Number.isNaN(uarg)) {
     throw new badArgumentError(name + ' must be a number');
   }
-
+  if (actualType === 'integer' && Number.isNaN(uarg)) {
```
bajtos: I think we need to check `Number.isInteger` here, shouldn't we?

lib/shared-method.js
bajtos: Can we change/simplify this condition please? I find it difficult to understand correctly.

```js
return targetType === 'integer' && Number.isInteger(val) ? 'integer' : 'number';
```

Also please add a unit-test to describe how `number` value `2.0` is handled.

test/rest.test.js
```diff
@@ -2253,6 +2276,8 @@ describe('strong-remoting-rest', function() {
   });
 
   it('rejects multi-item array passed to a number argument', function(done) {
+    /* it rejects multi-item array passed to integer argument
+     as it coerces to number type */
```
bajtos: Is this comment pointing out a missing unit-test?

test/shared-method.test.js
```diff
@@ -146,6 +146,50 @@ describe('SharedMethod', function() {
       });
     });
 
+    it('returns 400 when integer argument is float number',
+      function(done) {
+        var method = givenSharedMethod({
+          accepts: [{ arg: 'num', type: 'integer' }],
+        });
+
+        method.invoke('ctx', { num: NaN }, function(err) {
```
bajtos: `NaN` is not a flat number, please use a real float value here, e.g. `1.2`.

lib/shared-method.js
bajtos: > Also please add a unit-test to describe how number value 2.0 is handled.

Ah, is this covered by "treats number argument of type x.0 as number"?

bajtos:
@gunjpan Looks mostly good 👍 Please add a unit-test for "array of integers", address my comments above and fix test failures. I think adding support for "integer" should not be a breaking change, i.e. all existing tests should keep passing.

test/rest.test.js
gunjpan: yes, IIRC. As it will be an exact similar code block, will include the test case for `integer`

test/shared-method.test.js
gunjpan: My bad, I forgot to change after copying from the other case.

lib/shared-method.js
gunjpan: Yes, that's correct.

lib/shared-method.js
gunjpan: No, as there is already a check for the data type at [line#352](https://github.com/strongloop/strong-remoting/pull/323/files#diff-157ee1a11c353da754b459ce1fe0bc0dR352). This is for `NaN` values only.

lib/shared-method.js
```diff
@@ -576,12 +592,36 @@ SharedMethod.toResult = function(returns, raw, ctx) {
...
+//     var result = x;
+//     if (targetType === 'integer') {
+//       result = convertToIntegerTargetType(x);
+//       this.update(result);
+//     }
+//     return result;
+//   });
+//   return data;
+// }
+
```
gunjpan: Experimental

test/rest.test.js
```diff
@@ -950,29 +950,85 @@ describe('strong-remoting-rest', function() {
...
+            {
+              accepts: { arg: 'arg', type: ['number'] },
+              returns: { arg: 'data', type: ['integer'] },
+              http: { method: 'POST' },
+            });
+
+          request(app).post(method.url)
+            .send({ arg: [1.5, 2] })
+            .expect(200, { data: [3] })
+            .end(done);
```
gunjpan: This test-case is failing ATM, thus skipped.

bajtos:
@gunjpan there is one more thing we will need to do as part of fixing the issue, and that's adding new tests to rest-coercion integration tests in `test/rest-coercion`. I slightly prefer to do it as part of this patch, to ensure we don't miss any edge case here, but I won't mind if you decide to do that in a follow-up pull request.

lib/shared-method.js
bajtos: > No, as there is already a check for the data type at line#352.

Is my understanding correct that for non-integer numeric value, `actualType` will be `number`, and thus we will reject the value on the grounds that `number` is not `integer`? That makes sense.

Please add a unit-test to cover this case.

lib/shared-method.js
```diff
@@ -551,21 +558,34 @@ SharedMethod.toResult = function(returns, raw, ctx) {
...
-  function convert(val) {
-    switch (SharedMethod.getType(val)) {
+  function convert(val, targetType) {
+    var targetTypeIsArray = Array.isArray(targetType) &&
+      targetType.length === 1;
+
+    if (targetTypeIsArray && Array.isArray(val)) {
+      var arrayTargetType = targetType[0];
+      return val.map(function(arg, ix) {
+        return convert(arg, arrayTargetType);
```
bajtos: I think this is a breaking change, because we were not converting date/buffer types in an array before. I.e. array items were serialized using default `JSON.serialize` convention before, this convention is different from what strong-remoting implements. See https://github.com/strongloop/loopback/issues/1990 and https://github.com/strongloop/loopback/issues/1907 for more context.

I agree the new behaviour is better, because it's what most people would expect, but I am not sure if we can back-port such change to 2.x.

I am proposing to rework `convert` method to ignore all arrays except "array of integers", in order to preserve backwards compatibility, and then address the issue of encoding of array items later.

@gunjpan @ritch thoughts?

bajtos: Clarification of what I meant by my comment:

```
integer - do convert
date - do convert
buffer - do convert
array of integers - do convert
array of dates - no conversion
array of buffers - no conversion
```

gunjpan: @bajtos: Yeah, I agree with 
> rework convert method to ignore all arrays except "array of integers"

lib/shared-method.js
```diff
@@ -566,20 +572,20 @@ SharedMethod.toResult = function(returns, raw, ctx) {
...
+    var targetTypeIsIntegerArray = Array.isArray(targetType) &&
+      targetType.length === 1 && targetType[0] === 'integer';
 
-    if (targetTypeIsArray && Array.isArray(val)) {
+    if (targetTypeIsIntegerArray && Array.isArray(val)) {
       var arrayTargetType = targetType[0];
-      return val.map(function(arg, ix) {
-        return convert(arg, arrayTargetType);
+      return val.map(function(intVal) {
+        return convert(intVal, arrayTargetType);
```
bajtos: Should you forward `name` arg to `convert()` call here?

bajtos: Sorry, I meant `argName`.

lib/shared-method.js
```diff
@@ -596,14 +602,19 @@ SharedMethod.toResult = function(returns, raw, ctx) {
...
-      case 'number':
+      default:
         if (targetType === 'integer') {
-          if (!Number.isInteger(val)) return parseInt(val, 10);
+          if (!Number.isInteger(val)) {
+            var message = util.format(
+              'Invalid return value for argument \'%s\' of type ' +
+              '\'%s\': %s. Received type was %s.',
+              argName, targetType, val, typeof val);
+            throw new internalServerError(message);
```
bajtos: `internalServerError` already calls `new`, no need to call it again here.

test/rest-coercion/jsonform-integer.suite.js
```diff
@@ -0,0 +1,75 @@
...
+      // Empty values
+      [EMPTY_BODY, undefined],
+      [{ arg: null }, 0], // should be: null
+
+      // Valid values
+      [{ arg: 0 }, 0],
+      [{ arg: 1 }, 1],
+      [{ arg: -1 }, -1],
+
+      // Integers larger than MAX_SAFE_INTEGER get trimmed
```
bajtos: IMO, integers larger than MAX_SAFE_INTEGER should be rejected with 400 Bad Request. We can do that change outside of this pull request if you prefer, but the comment should be changed in such case.

test/rest-coercion/jsonform-integer.suite.js
```diff
@@ -0,0 +1,75 @@
...
+      [{ arg: 2343546576878989879789 }, 2.34354657687899e+21],
+      [{ arg: -2343546576878989879789 }, -2.34354657687899e+21],
+
+      // Scientific notation works
+      [{ arg: 1.234e+30 }, 1.234e+30],
+      [{ arg: -1.234e+30 }, -1.234e+30],
+
+      // Integer-like string values should not trigger ERROR_BAD_REQUEST
```
bajtos: Ideally, I would expect that JSON request body `{ arg: '0' }` triggers 400 Bad request when `arg` is expected to be a number. The test suite for numbers ([link](https://github.com/strongloop/strong-remoting/blob/27db434b7b252efbfea7128142e5b49a49eb16a5/test/rest-coercion/jsonform-number.suite.js#L53)) says "Number-like string values should trigger ERROR_BAD_REQUEST", I don't see why integers should behave differently.

test/rest-coercion/jsonform-integer.suite.js
```diff
@@ -0,0 +1,75 @@
...
+      [{ arg: '0' }, 0],
+      [{ arg: '1' }, 1],
+      [{ arg: '-1' }, -1],
+      [{ arg: '2343546576878989879789' }, 2.34354657687899e+21],
+      [{ arg: '-2343546576878989879789' }, -2.34354657687899e+21],
+      [{ arg: '1.234e+30' }, 1.234e+30],
+      [{ arg: '-1.234e+30' }, -1.234e+30],
+
+      // All other non-integer values should trigger ERROR_BAD_REQUEST
+      [{ arg: '1.2' }, ERROR_BAD_REQUEST],
```
bajtos: You are missing few test cases here for values that should be rejected:

```js
{ arg: 1.2 },
{ arg -1.2 },
```

test/rest-coercion/urlencoded-integer.suite.js
```diff
@@ -0,0 +1,73 @@
...
+      [EMPTY_QUERY, undefined],
+      ['arg', 0], // should be: undefined
+      ['arg=', 0], // should be: undefined
+
+      // Valid values
+      ['arg=0', 0],
+      ['arg=1', 1],
+      ['arg=-1', -1],
+
+      // Numbers larger than MAX_SAFE_INTEGER get trimmed
```
bajtos: Should be rejected, see https://github.com/strongloop/strong-remoting/pull/323/files/e8649ef1693273e0c8fa698fa7c7bfc70d422552..e066d0e76c608e014d312f6efac0281c08392bc0#r72065917

test/shared-method.test.js
```diff
@@ -202,8 +203,29 @@ describe('SharedMethod', function() {
...
-    });
 
+      it('returns 500 for non-integer return value if type: `integer`',
+        function(done) {
+          var method = givenSharedMethod(
+            function(cb) {
+              cb(null, 3.141);
+            },
+            {
+              returns: [
```
bajtos: Nitpick: L214-L216 can be abbreviated to

```js
returns: { arg: 'value', type: 'integer' }
```

lib/shared-method.js
gunjpan: yes, it should.

gunjpan:
@bajtos : Previous comments addressed. PTAL at the latest commit. Thanks.

lib/shared-method.js
```diff
@@ -701,3 +716,15 @@ SharedMethod.prototype.addAlias = function(alias) {
...
+
+// Include support for following Number methods to support Node v.0.10.x
+Number.isInteger = Number.isInteger || function(value) {
+  return typeof value === 'number' &&
+    isFinite(value) &&
+    Math.floor(value) === value;
+};
+
+Number.isSafeInteger = Number.isSafeInteger || function(value) {
+  Number.MAX_SAFE_INTEGER = Number.MAX_SAFE_INTEGER || Math.pow(2, 53) - 1;
```
bajtos: Please move this line outside of `Number.isSafeInteger` function, no need to run it every time `isSafeInteger` is called.

test/rest-coercion/jsonform-integer.suite.js
```diff
@@ -40,24 +40,26 @@ module.exports = function(ctx) {
...
       [{ arg: '1' }, 1],
       [{ arg: '-1' }, -1],
-      [{ arg: '2343546576878989879789' }, 2.34354657687899e+21],
-      [{ arg: '-2343546576878989879789' }, -2.34354657687899e+21],
-      [{ arg: '1.234e+30' }, 1.234e+30],
-      [{ arg: '-1.234e+30' }, -1.234e+30],
+      [{ arg: '2343546576878989879789' }, ERROR_BAD_REQUEST],
+      [{ arg: '-2343546576878989879789' }, ERROR_BAD_REQUEST],
+      [{ arg: '1.234e+30' }, ERROR_BAD_REQUEST],
+      [{ arg: '-1.234e+30' }, ERROR_BAD_REQUEST],
```
bajtos: Please add two more items: `'1.234e+3` and `'-1.234e+3'` (strings containing an integer using scientific notation)

bajtos:
Two more comments, you are almost there :)

bajtos:
LGTM. Please squash the commits and wait for CI build to turn green before landing.

gunjpan:
@slnode test please

gunjpan:
@slnode test please

gunjpan:
@bajtos: Landed. Do we need to backport it to 2.x?

Issue #324, opened 2016-07-21
Querystring behaviour is inconsistent when integrating other Loopback APIs


horiaradu:
Loopback has a particular interface for REST APIs. It allows you to write a query like so:

`Model.find({where: {id: {inq: [1, 2, 3]}}})`

However, this query is also valid: `Model.find({where: {id: {inq: []}}})`, but it will obviously not return any results.

In this case, the generated query looks like this:

```
http://localhost:3010/api/model?filter=%7B%22where%22%3A%20%7B%22id%22%3A%20%7B%22inq%22%3A%20%5B%5D%7D%7D%7D
```

and decoded, like this:

```
http://localhost:3010/api/model?filter={"where": {"id": {"inq": []}}}
```

The problem is inside [http-invocation.js](https://github.com/strongloop/strong-remoting/blob/master/lib/http-invocation.js), line: 215. 

```
// query is an object: query = {where: {id: {inq: []}}}
qs.stringify(query) //this will return an empty string
```

Querystring behaviour (from node REPL):

```
> qs.stringify({where: {id: {inq: [1]}}})
'where%5Bid%5D%5Binq%5D%5B0%5D=1'
> qs.stringify({where: {id: {inq: []}}})
''
>
```

If you are integrating another loopback API, when you do the query with an empty array, you expect to get back an empty collection of items, but instead, because of the behaviour of querystring, you get back **all** the items.

See related issues:
- on querystring: https://github.com/nodejs/node-v0.x-archive/issues/8829
- on datasource-juggler: https://github.com/strongloop/loopback-datasource-juggler/issues/1006

BoLaMN:
references #320

richardpringle:
Hey @horiaradu, what version of `loopback`, `strong-remoting`, and `loopback-datasource-juggler` are you using?

I just tried to reproduce the issue and it was working as expected. Either that, or I don't entirely understand the issue. You're saying that when you use an empty array, it returns all the results instead of none?

BoLaMN:
The issue I posted is still relevant in how querystring handles arrays throwing object ids on top just makes the whole scenario worse.

I was getting 413 errors due to the querystring being to large for nginx to process, check my referenced issue to find an example in tonicdev you can play with.

In the mean time I too have forked and introduced a similar patch.

cheers

horiaradu:
@richardpringle 

```loopback```: 2.29.1
```strong-remoting```: 2.28.1
[...]

PR #325, opened 2016-07-21
Stringify query params which are objects as JSON

horiaradu:
Stringify query params which are objects in a way in which empty ararys
are preserved instead of removed (default querystring implementation).

fix: https://github.com/strongloop/strong-remoting/issues/324

### DIFF
--- a/lib/http-invocation.js
+++ b/lib/http-invocation.js
@@ -128,7 +128,7 @@ HttpInvocation.prototype._processArg = function(req, verb, query, accept) {
             // From the query string
             if (val !== undefined) {
               query = query || {};
-              query[name] = val;
+              query[name] = serializeQueryStringValue(val, accept);
             }
             break;
           case 'header':
@@ -148,7 +148,7 @@ HttpInvocation.prototype._processArg = function(req, verb, query, accept) {
     // default to query string for GET
     if (val !== undefined) {
       query = query || {};
-      query[name] = val;
+      query[name] = serializeQueryStringValue(val, accept);
     }
   } else {
     // default to storing args on the body for !GET
@@ -158,6 +158,15 @@ HttpInvocation.prototype._processArg = function(req, verb, query, accept) {
 
   return query;
 };
+
+function serializeQueryStringValue(val, accept) {
+  if ((accept.type === 'object' || accept.type === 'string') &&
+    typeof val === 'object') {
+    return JSON.stringify(val);
+  } else {
+    return val;
+  }
+}
 
 /**
  * Build args object from the http context's `req` and `res`.
--- a/test/http-invocation.test.js
+++ b/test/http-invocation.test.js
@@ -193,6 +193,133 @@ describe('HttpInvocation', function() {
       inv.transformResponse(res, body, cb);
     }
   });
+
+  describe('createRequest', function() {
+    it('creates a simple request', function() {
+      var inv = givenInvocationForEndpoint(null, []);
+      var expectedReq = { method: 'GET',
+        url: 'http://base/testModel/testMethod',
+        protocol: 'http:',
+        json: true,
+      };
+      expect(inv.createRequest()).to.eql(expectedReq);
+    });
+
+    it('makes primitive type arguments as query params', function() {
+      var accepts = [
+        { arg: 'a', type: 'number' },
+        { arg: 'b', type: 'string' },
+      ];
+      var aValue = 2;
+      var bValue = 'foo';
+      var inv = givenInvocationForEndpoint(accepts, [aValue, bValue]);
+      var expectedReq = { method: 'GET',
+        url: 'http://base/testModel/testMethod?a=2&b=foo',
+        protocol: 'http:',
+        json: true,
+      };
+      expect(inv.createRequest()).to.eql(expectedReq);
+    });
+
+    it('makes an array argument as a query param', function() {
+      var accepts = [
+        { arg: 'a', type: 'object' },
+      ];
+      var aValue = [1, 2, 3];
+      var inv = givenInvocationForEndpoint(accepts, [aValue]);
+      var expectedReq = { method: 'GET',
+        url: 'http://base/testModel/testMethod?a=' + encodeURIComponent('[1,2,3]'),
+        protocol: 'http:',
+        json: true,
+      };
+      expect(inv.createRequest()).to.eql(expectedReq);
+    });
+
+    it('keeps an empty array as a query param', function() {
+      var accepts = [
+        { arg: 'a', type: 'object' },
+      ];
+      var aValue = [];
+      var inv = givenInvocationForEndpoint(accepts, [aValue]);
+      var expectedReq = { method: 'GET',
+        url: 'http://base/testModel/testMethod?a=' + encodeURIComponent('[]'),
+        protocol: 'http:',
+        json: true,
+      };
+      expect(inv.createRequest()).to.eql(expectedReq);
+    });
+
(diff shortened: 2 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

horiaradu:
Any updates on this one?

richardpringle:
ok to test

richardpringle:
none of the tests you added are passing on my machine

richardpringle:
@slnode test please

richardpringle:
@horiaradu, the change that was made in this PR only affects the behavior of a `GET` request, where the source of the argument has not been defined, ie the default `GET` request behavior. Shouldn't we put this change in for any argument sourced from the query string?

Also, what happens if `typeof accept.type === 'object'` and `typeof val === 'object'`? You're still coercing the object into a string. Is your argument that any argument coming from the query should be a string? I'm don't necessarily disagree, I just want to clarify.

horiaradu:
@richardpringle 

What I want to accomplish is to serialize objects as strings when they need to go into the query string.

I only affected the GET request since I encountered this bug when I connected to a loopback API (as described [here](https://github.com/strongloop/strong-remoting/issues/324)) and only GET requests were used. I don't have enough knowledge of strong-remoting to know how the other requests are affected, but I can gladly try to help out in order to make this more generic.

richardpringle:
We should definitely make sure that the change is uniform and that any object parsed from the query string is coerced properly. 

If you could test the expected behaviour is the same for multiple verbs (GET and POST for example), that would be ideal.

horiaradu:
I will replicate the tests for other verbs as well.

thaiat:
any reason why this is not merged ?

gunjpan:
@thaiat 
>any reason why this is not merged ?

Looks like there's pending item on this PR. Also, has merge conflicts.

@horiaradu : Can you please advise if you were able to finish:
> I will replicate the tests for other verbs as well.

Please resolve merge conflicts as well.

horiaradu:
I will resolve the conflicts asap.

horiaradu:
@gunjpan I've updated the PR.

* merge conflicts resolved
* added a couple of specs for POST methods

Check it out and let me know if I have to fix other things.

gunjpan:
@slnode ok to test

thaiat:
this should be ported back to latest 2.x version of strong-remoting as well

horiaradu:
I can take care of that after this PR is merged.

thaiat:
@horiaradu i needed that in my code asap so i did it there in case that helps you #377

gunjpan:
Looks like CI is stuck. Restarting tests.

gunjpan:
@slnode test please

horiaradu:
I would want to include the functionality with the unit tests as well so I
will just cherry pick the commit.

horiaradu:
I see that this PR appeared on the same topic: https://github.com/strongloop/strong-remoting/pull/378

Can somebody take care of this PR? It's been here for a while now...

Amir-61:
@horiaradu I see @gunjpan is taking care of reviewing this PR. However before that you need to resolve merge conflicts. If this PR gets landed we can close https://github.com/strongloop/strong-remoting/pull/378

Thanks!

horiaradu:
I don't see any conflicts...

Amir-61:
>I don't see any conflicts...

Oh it says:
> This branch is out-of-date with the base branch

I believe you need to rebase it on top of master.

horiaradu:
done.

Amir-61:
@gunjpan could you PTAL.

lib/http-invocation.js
```diff
@@ -148,7 +148,13 @@ HttpInvocation.prototype._processArg = function(req, verb, query, accept) {
...
     if (val !== undefined) {
       query = query || {};
-      query[name] = val;
+
+      if ((accept.type === 'object' || accept.type === 'string') &&
+        typeof val === 'object') {
+        query[name] = JSON.stringify(val);
+      } else {
+        query[name] = val;
+      }
```
gunjpan: @horiaradu : This needs to be applied for `query` case in here as well:
https://github.com/strongloop/strong-remoting/blob/master/lib/http-invocation.js#L127-L132

Better solution would be to extract it in a common function.

test/http-invocation.test.js
```diff
@@ -193,6 +193,182 @@ describe('HttpInvocation', function() {
...
   });
+
+  describe('createRequest', function() {
+    it('should create a simple request', function() {
+      var method = givenSharedStaticMethod();
+      method.getEndpoints = function() {
+        return [createEndpoint()];
+      };
+      var inv = givenInvocation(method, { ctorArgs: [], args: [], baseUrl: 'http://base' });
+
```
gunjpan: line 199:203 could be extracted to a single function as I see them being repeated.

gunjpan:
@horiaradu : I found pull #378 more elegant and generalized. 

@bajtos : Thoughts?

horiaradu:
@gunjpan The original reason this PR was made was to handle the **filter** query parameter. There, you have a full object with nested values which can be null / empty array.

#378 only handles the first level params and will fail if the null / empty array appears at a deeper level. You would need a recursive object traversal solution. Plus it does not add any tests and I think the unit tests are critical for a library such as this.

I've made the changes you suggested.

bajtos:
> @horiaradu : I found pull #378 more concise and generalized, because `JSON.stringify()` would retain numbers as numbers: From node REPL: 
> ```Console
> console.log(JSON.stringify({a: 2}))
{"a":2}
```

> @bajtos : Thoughts? 

I think #378 goes in a wrong direction. What happens when the argument value is a string? I think we will end up with `?key="value"` in the query, which is definitely not wanted.

```
> JSON.stringify('text')
'"text"'
```

This patch looks good to me as it is now, I think the only remaining step is to squash the commits to a single one with a good commit message ([guidelines](http://loopback.io/doc/en/contrib/git-commit-messages.html)). I'll leave the final call to @gunjpan.

**Comment by gunjpan** [comment]
@bajtos : I see, thank you for clarifying the case of `string`

@horiaradu : PTAL at @bajtos' comments above. I'll further review your edits today. Thank you.

**Comment by horiaradu** [comment]
done

File: test/http-invocation.test.js [hunk 3]
```diff
@@ -193,6 +193,146 @@ describe('HttpInvocation', function() {
       inv.transformResponse(res, body, cb);
     }
   });
+
+  describe('createRequest', function() {
+    it('should create a simple request', function() {
```
gunjpan: @horiaradu : Please change all the `it` cases to reflect **imperative mood** as described in the Code style guidelines [here](http://loopback.io/doc/en/contrib/style-guide.html#test-naming).

test/http-invocation.test.js
```diff
@@ -193,6 +193,146 @@ describe('HttpInvocation', function() {
...
+      var inv = givenInvocationForEndpoint(null, []);
+
```
gunjpan: Please remove extra line here and
group together `var` declarations, for example:
```javascript
 var a = 'Hello';
 var b = 'How are you';

 function printMessage(a, b){
   console.log(a + ', ' + b);
}

```
 
Same applies to other it blocks

horiaradu:
@gunjpan check it out.

I will backport this to 2.X and create a PR as soon as this is merged.

gunjpan:
@horiaradu : Landed. Please backport it against `2.x`, and ping me for review. Thank you :)

PR #333, opened 2016-08-12
Remove the usage of deprecated methods

Amir-61:
* Remove the usage of deprecated methods in shared-class.test.js

Also it needs to be back-ported.

/to: @bajtos

### DIFF
--- a/test/shared-class.test.js
+++ b/test/shared-class.test.js
@@ -9,6 +9,7 @@ var expect = require('chai').expect;
 var SharedClass = require('../lib/shared-class');
 var factory = require('./helpers/shared-objects-factory.js');
 var RemoteObjects = require('../');
+function NOOP() {};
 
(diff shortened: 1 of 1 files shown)


test/shared-class.test.js
```diff
@@ -143,12 +144,13 @@ describe('SharedClass', function() {
       SomeClass.prototype.myMethod = function() {};
       var METHOD_NAME = 'myMethod';
       sc.defineMethod(METHOD_NAME, {
+        isStatic: true,
```
Amir-61: I could also use `isStatic: false` too, but in this case Line 153 needs to be changed to:

```
expect(sc.findMethodByName('prototype.myMethod').accessType).to.eql('READ');
```

bajtos:
Nice, :shipit:

PR #336, opened 2016-08-16
Add new API for finding/disabling remote methods

bajtos:
This is a back-port of #275 excluding the deprecation warnings.

If we included deprecation warnings, then I think tools like loopback-swagger and loopback-sdk-angular will start reporting these deprecations, because IIRC they are still using the old API. Thus I think it's better to buy us more time by not deprecating the old methods yet.

@0candy please review
cc @Amir-61 @davidcheung


0candy:
LGTM. Needs rebasing.

Issue #312, opened 2016-06-10
Fix coercion of input arguments (HTTP/REST)


bajtos:
Based on the integration test suite written in #304, I am proposing to make the following changes/fixes in the way how we coerce input arguments.

I think most (if not all) of this changes can be considered as backwards-compatible fixes to be landed on 2.x too.

@STRML @ritch thoughts?

---

Terminology:

 - `json body` - the argument is set to the full request body which is JSON encoded
 - `json form` - the argument is read from a property in the JSON-encoded request body
 - `urlencoded` - the argument is read either from query string or from urlencoded request body

## Tighten validation of required arguments

 - `json body - any - required` should reject `null` value
 - `json form - array - required` should reject empty request and `null` value

 - `json form - * - required` should reject empty request, `null` value
   and empty string value

 - `urlencoded - *` should reject empty query, `?arg`, `?arg=`, `?arg=null`

## Reject scalar values for array type

 - `json body - array of *`
 - `json form - array - required`

## Do not coerce missing value to empty array

This may be possibly controversial. Should we introduce a config option
to control this behaviour?

 - `json form - array of *` should convert empty body to `undefined`
 - `json form - array of *` should convert `null` to `null`
 - `urlencoded - array of *` should convert `?`, `?arg`, `?arg=` to `undefined`

## Do not coerce values from JSON-encoded sources

 - Reject requests sending array items of wrong type
  - `json body - array of *`
  - `json form - array of *`

 - Do not coerce array items from string to number, boolean, etc.
   - `json body - array of any`
   - `json form - array of any`

 - Reject requests providing value of wrong type
   - `json form - boolean` (reject strings, numbers, etc.)
   - `json form - date` (reject booleans, large numbers, non-date strings, etc.)
   - `json form - number` (reject strings, booleans, arrays, etc.)
   - `json form - object` (reject strings, booleans, numbers)
   - `json form - string` (reject booleans, numbers, objects, arrays, etc.)

## Allow `null` as a value for object type

Convert `null` input to `null` value:

 - `json body - object - optional`
 - `json form - *`
 - `urlencoded - *`

## Treat missing argument as `undefined`

Set argument to `undefined` when not set:

 - `json form - *`
 - `urlencoded - *` - treat `?arg` as `undefined`

## Don't coerce too large numbers

 - `urlencoded - any` - keep `?arg=2343546576878989879789` as string value

## Recognize scientific notation when parsing strings

[...]

PR #343, opened 2016-08-31
[SEMVER-MAJOR] Rework coercion of input arguments

bajtos:
Simplify the design of input argument handling.

 - It's up to the Context and/or the Adapter to convert or coerce
   input arguments as needed.

 - The only responsibility of SharedMethod is to validate input
   arguments (check that they match the contract in "accepts").

 - Define two types of conversion/coercion:
   a) When the input argument value comes from a source that preserves
      type information (typically JSON), then the argument should be
      converted with minimal coercion.
   b) When the input argument value comes from a "sloppy" source
      (typically string-encoded source like a query string), then
      coercion rules should be applied.

Replace "Dynamic" API, which was relying on global singleton registry,
with a new concept of a "TypeRegistry" and a type "Converter".
The new registry is local to each RemoteObjects instance.

Each type converter provides three methods:
  - `fromTypedValue` to handle values coming from JSON
  - `fromSloppyValue` to handle values coming e.g. from query string
  - `validate` which is called by SharedMethod to validate argument
    values

To prevent unnecessary try/catch blocks with the associated performance
costs, validation/conversion errors are not thrown but returned via
function return value instead.

The feature where one could encode all input arguments in a single
(query string) parameter "args" is removed.

This patch fixes many subtle bugs and changes the way how edge case
values are treated. The general approach is to make both conversion and
coercion more strict. When we are not sure how to treat an input value,
we rather return HTTP 400 Bad Request than coerce the value incorrectly.

Most notable breaking changes:

 - `null` value is accepted only for "object", "array" and "any"
 - Empty string is coerced to undefined to support ES6 default arguments
 - JSON requests providing scalar values for array-typed argument are
   rejected
 - Empty value is not converted to an empty array
 - Array values containing items of wrong type are rejected. For
   example, an array containing a string value is rejected when
   the input argument expects an array of numbers.
 - Array items from JSON requests are not coerced. For example,
   `[true]` is no longer coerced to `[1]` for number arrays,
   and the request is subsequently rejected.
 - Deep members of object arguments are no longer coerced. For example,
   a query like `?arg[count]=42` produces `{ count: '42' }` now.
 - "any" coercion preserves too large numbers as a string, to prevent
   losing precision.
 - Boolean types accepts only four string values:
    'true', 'false', '0' and '1'
   Values are case-insensitive, i.e. 'TRUE' and 'FaLsE' are work too.
 - Date type detects "Invalid Date" and rejects such requests.

Last but not least, conversion/validation error messages no longer
include input argument values, in order to prevent reflection-based
attacks.

@gunjpan @richardpringle @deepakrkris Please review.
@ritch @STRML PTAL too, I would like to hear your opinion on this change.
@0candy @davidcheung FYI

This patch is loosely inspired by https://github.com/strongloop/strong-remoting/pull/264.

The change-set is quite large and it may be too difficult to review all details. Here are the most important parts to check:
 - the overal design - converters, type registry, what work is done by Context vs. SharedMethod
 - changes in rest-coercion test suite - is there anything surprising?
 - changes in seemingly unrelated files (lib/context-base, lib/http-context, lib/http-invocation)

Connect to #312
Connect to strongloop-internal/scrum-loopback#974
Connect to strongloop-internal/scrum-loopback#885

TODO for @bajtos:
 - [x] Add an entry to 3.0-RELEASE-NOTES
 - [x] Update loopback@master to work with the new version - https://github.com/strongloop/loopback/pull/2696
 - [x] Ensure loopback-connector-remote works with the new version - https://github.com/strongloop/loopback-connector-remote/pull/58


bajtos:
Cross-posting @STRML comment from https://github.com/strongloop/strong-remoting/issues/312#issuecomment-225174905

> If you're accepting an array input for a field, I don't see any semantic difference between `[]` (no items) and `null` (no items). But `[]` is much more convenient as you can simply iterate.

Here is an example where the difference is important: let's say we have a method `find` accepting an argument called `extraFields`.  When `extraFields` is present, the method adds the given fields to the response data. When `extraFields` is not present, all extra fields are included.

Having said that, I see your point of view @STRML. I think we can achieve what you are asking for by supporting a `default` value in argument definition:

```js
accepts: {
  arg: 'numbers', type: ['number'], default: []
}
```

Let's leave the support for default values out of scope of this patch though.

bajtos:
Also: for long time, I though we can land (some of) these changes to 2.x in such way that backwards compatibility is preserved. I changed my mind after I checked the effect of #264 on the rest-coercion test suite, there are too many changes where I am concerned about the impact on existing applications. Our current plan is to release loopback@3 by the end of September, therefore I think it will be easier for all of us if this patch is landed on 3.0 only.

richardpringle:
@slnode test please

bajtos:
I think the CI results cannot be trusted very much. First of all, this patch is targeting 3.0 release line, which is used in very few downstream projects. Secondly, `loopback` requires changes in https://github.com/strongloop/loopback/pull/2696 to pass the CI, I am not sure if CIS will pick the right branch in loopback when running downstream tests of this pull request.

richardpringle:
@bajtos,
The `strong-remoting` build itself was not passing before, that's why I re-triggered the test. It seems to be passing now.

3.0-RELEASE-NOTES.md
```diff
@@ -113,3 +113,66 @@ An example of an error response when an array of errors was raised:
...
+ - Array items from JSON requests are not coerced. For example,
+   `[true]` is no longer coerced to `[1]` for number arrays,
+   and the request is subsequently rejected.
+ - Deep members of object arguments are no longer coerced. For example,
+   a query like `?arg[count]=42` produces `{ count: '42' }` now.
+ - "any" coercion preserves too large numbers as a string, to prevent
+   losing precision.
+ - Boolean types accepts only four string values:
+    'true', 'false', '0' and '1'
+   Values are case-insensitive, i.e. 'TRUE' and 'FaLsE' are work too.
```
richardpringle: ~~are~~ work too

richardpringle:
The review process could be much faster if this spread out into multiple commits.

lib/remote-objects.js
```diff
@@ -707,22 +708,90 @@ RemoteObjects.prototype.getScope = function(ctx, method) {
...
+ *     // use the val and ctx objects to return the concrete value
+ *     if (value === undefined || value === null)
+ *       return { value: value };
+ *
+ *     value = new MyType(value);
+ *     var error = this.validate(ctx, value);
+ *     return error ? { error: error } : { value: value };
+ *   },
+ *
+ *   fromSloppyValue: function(ctx, value) {
```
richardpringle: Maybe add a definition of a "sloppy value"?

lib/types/any.js
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    if (value.toLowerCase() === 'true')
+      return { value: true };
+
+    if (value.toLowerCase() === 'false')
+      return { value: false };
+
+    if (IS_FLOAT_REGEX.test(value) || IS_INT_REGEX.test(value)) {
+      var num = +value;
```
gunjpan: I believe, you're assigning a value here: `var num = value`. If that's the case, please correct. If not, I didn't get the point of using `+` there.

ritch:
```js
Ω node
> value = '123'
'123'
> +value
123
> Number(value)
123
```

`+value` is just a way of casting/coercing a string into a number... I prefer `Number(value)` or in this case... `parseInt(value)` / `parseFloat(value)`.

ritch:
Removing `Dynamic` will break a lot of things in LoopBack core.

lib/remote-objects.js
```diff
@@ -707,22 +708,90 @@ RemoteObjects.prototype.getScope = function(ctx, method) {
...
+ *
+ * ```
+ * remotes.defineObjectType('MyClass', function(data) {
+ *   return new MyClass(data);
  * });
  * ```
  *
- * **Note: the alias `remotes.convert()` is deprecated.**
+ * Under the hood, a converter is created that ensures the input data
+ * is an object (or sloppy value is coerced to an object) and calls
```
ritch: > sloppy value
I see this referenced in various places, and I am not sure what it means. 

Does this mean the raw value provided from `req.param()` or equiv?

lib/types/any.js
gunjpan: @ritch : Gotcha! thanks for clarifying.

bajtos:
@ritch 
> Removing `Dynamic` will break a lot of things in LoopBack core.

Not really. We use `Dynamic` in one place only and the upgrade is trivial - see https://github.com/strongloop/loopback/pull/2696/commits/826205997d0bf861d1b81ac2301417fa28d0962b.

lib/remote-objects.js
bajtos: Good point. I took the concept of "sloppy" values from #264 but on the second thought the name is not descriptive enough.

The sloppy conversion/coercion is invoked when the value comes from a string-only source like HTTP header or more importantly a query string parsed via `qs`. 

A sloppy value is one of:
 - a string
 - an array containing sloppy values
 - an object where property values are sloppy

I am tempted to rename `fromSloppyValue` to `fromStringValue`, but that would hide the fact that an array or an object can be passed in too.

@ritch @richardpringle Any suggestions for a better name?

I reworked the comments in 0b7df4fa2 to make this matter more clear, PTAL.

lib/remote-objects.js
bajtos: See https://github.com/strongloop/strong-remoting/pull/343#discussion_r77302252

> The sloppy conversion/coercion is invoked when the value comes from a string-only source like HTTP header or more importantly from a query string parsed via qs.

Should I expand this comment to make it more clear what is meant by sloppy values?

lib/context-base.js
```diff
@@ -5,18 +5,28 @@
...
  * A base class for all Context instances
  */
-function ContextBase(method) {
+function ContextBase(method, typeRegistry) {
+  // NOTE(bajtos) we are not asserting via "instanceof" to allow
+  // multiple copies of strong-remoting to cooperate together
+  assert(method && typeof method === 'object',
+    'method must be a SharedClass instance');
+  assert(typeRegistry && typeof typeRegistry === 'object',
+    'typeRegistry must be a TypeRegistry instance');
```
Amir-61: We have used them a lot in `dao.js` as well; this `assert(...)` things are so neat; however I just wonder if the assertion fails, does it dump the stack trace?

lib/context-base.js
```diff
@@ -29,3 +39,7 @@ ContextBase.prototype.getScope = function() {
     method.ctor ||
     method.sharedMethod && method.sharedMethod.ctor;
 };
+
+ContextBase.prototype.setReturnArgByName = function(name, value) {
+  // no-op
+};
```
Amir-61: Nitpick: Shouldn't we add an empty line at the end of file?

lib/context-base.js
```diff
@@ -29,3 +39,7 @@ ContextBase.prototype.getScope = function() {
     method.ctor ||
     method.sharedMethod && method.sharedMethod.ctor;
 };
+
+ContextBase.prototype.setReturnArgByName = function(name, value) {
```
Amir-61: Suggestion: I wonder whether if it will be useful if we move `function NOOP() {}` to the top of the file and call it wherever needed?

3.0-RELEASE-NOTES.md
```diff
@@ -113,3 +113,66 @@ An example of an error response when an array of errors was raised:
 ```
 
 Please see [Related code change](https://github.com/strongloop/strong-remoting/pull/302) here, and new [`strong-error-handler` here](https://github.com/strongloop/strong-error-handler/).
+
+## Type converters replace `Dynamic` API
+
+The `Dynamic` component was removed in favour of type converters and a
+registry.
+
+The following API was removed:
```
Amir-61: The following APIs were removed ?

lib/http-context.js
```diff
@@ -171,55 +177,36 @@ HttpContext.prototype.buildArgs = function(method) {
...
+
+    debug('arg %j: %s converted %j to %j',
+      name, doSloppyCoerce ? 'sloppy' : 'typed', val, result);
+
+    var isValidResult = typeof result === 'object' &&
+      ('error' in result || 'value' in result);
+    if (!isValidResult)  {
+      throw new (assert.AssertionError)({
+        message: 'Type conversion result should have "error" or "value" property. ' +
+          'Got ' + JSON.stringify(result) + ' instead.',
```
Amir-61: Shouldn't this be globalized?

lib/number-checks.js
```diff
@@ -0,0 +1,15 @@
+// To support Node v.0.10.x
```
Amir-61: Shouldn't we add IBM header for this newly created file?

lib/number-checks.js
```diff
@@ -0,0 +1,15 @@
...
+var number = module.exports = {
+  MAX_SAFE_INTEGER: Number.MAX_SAFE_INTEGER || 9007199254740991,
```
Amir-61: Just a suggestion; wouldn't you thing the following would be neater and makes sense to the reader:
```
MAX_SAFE_INTEGER: Number.MAX_SAFE_INTEGER || Math.pow(2, 53) - 1,
```

lib/number-checks.js
```diff
@@ -0,0 +1,15 @@
...
+    return typeof value === 'number' &&
+      isFinite(value) &&
+      Math.floor(value) === value;
+  },
+  isSafeInteger: Number.isSafeInteger || function(value) {
+    return number.isInteger(value) &&
+      value >= number.MIN_SAFE_INTEGER &&
+      value <= number.MAX_SAFE_INTEGER;
+  },
+};
```
Amir-61: Empty line at the end of file?

lib/shared-method.js
```diff
@@ -274,7 +272,7 @@ SharedMethod.prototype.invoke = function(scope, args, remotingOptions, ctx, cb)
     }
     return retval;
   } catch (err) {
-    debug('error caught during the invocation of %s', this.name);
+    debug('error caught during the invocation of %s', this.name, err);
```
Amir-61: Where is `err` being used? Are you missing `%j`?

lib/shared-method.js
```diff
@@ -304,86 +302,27 @@ function escapeRegex(d) {
...
-      name, targetType, uarg, typeof uarg);
-    throw badArgumentError(message);
-  }
 
   // Verify that a required argument has a value
-  // FIXME(bajtos) "null" should be treated as no value too
-  if (actualType === 'undefined') {
-    if (desc.required) {
+  if (desc.required) {
+    var argIsNotSet = uarg === null || uarg === undefined || uarg === '';
```
Amir-61: How about simplifying to:
```
var argIsNotSet = uarg == null  || uarg === '';
```
thought?

lib/type-registry.js
```diff
@@ -0,0 +1,118 @@
...
+  assert(typeof converter.validate === 'function',
+    'converter.validate must be a function');
+
+  typeName = typeName.toLowerCase();
+
+  if (typeName in this._types) {
+    if (this._options.warnWhenOverridingType !== false)
+      g.warn('Warning: overriding remoting type %s', typeName);
+    else
+      debug('Warning: overriding remoting type %s', typeName);
```
Amir-61: Actually I love this ^ syntax; I mean skipping curly braces for one line `if/else`, but is this what tea follows? It reminds me of coffee script too :-)

Amir-61:
@bajtos I reviewed almost one thirds of your pull request tonight; I'll try to review the rest to unblock you :-)

lib/type-registry.js
```diff
@@ -0,0 +1,118 @@
...
+
+  this.registerType(typeName, converter);
+};
+
+TypeRegistry.prototype.getConverter = function(type) {
+  assert(typeof type === 'string' || Array.isArray(type),
+    'type must be either an array or a string.');
+
+  if (type === 'array')
+    type = ['any'];
```
Amir-61: ditto

lib/types/any.js
```diff
@@ -0,0 +1,60 @@
...
+var IS_INT_REGEX = /^\-?(?:[0-9]|[1-9][0-9]*)$/;
+var IS_FLOAT_REGEX = /^\-?([0-9]+)?\.[0-9]+$/;
+
+module.exports = {
+  fromTypedValue: function(ctx, value) {
+    return { value: value };
+  },
+
+  fromSloppyValue: function(ctx, value) {
+    if (value === 'null' || value === null)
```
Amir-61: Question: If `value` is a string of null; is it always lowercase null and not like `NULL`?

lib/types/array.js
```diff
@@ -0,0 +1,157 @@
...
+var g = require('strong-globalize')();
+
+module.exports = ArrayConverter;
+
+function ArrayConverter(itemType) {
+  this._itemType = itemType;
+}
+
+ArrayConverter.prototype.fromTypedValue = function(ctx, value) {
+  if (value === undefined || value === null)
```
Amir-61: ditto; how about:
```
if (value == null)
```

lib/types/array.js
```diff
@@ -0,0 +1,157 @@
...
+    if (itemResult.error)
+      return itemResult;
+    items[ix] = itemResult.value;
+  }
+  return { value: items };
+};
+
+ArrayConverter.prototype.fromSloppyValue = function(ctx, value) {
+  if (value === undefined || value === '') {
+    // undefined was chosen so that it plays well with ES6 default parameters.
```
Amir-61: 👍

lib/types/array.js
```diff
@@ -0,0 +1,157 @@
...
+  if (!looksLikeJsonArray)
+    return null;
+
+  // If it looks like a JSON array, try to parse it.
+  try {
+    var result = JSON.parse(value);
+    debug('parsed %j as JSON: %j', value, result);
+    return this.fromTypedValue(ctx, result);
+  } catch (ex) {
+    debug('Cannot parse array value %j. %s', value, ex);
```
Amir-61: Question: Is `ex` object? Shouldn't we use `%j`?

lib/types/array.js
```diff
@@ -0,0 +1,157 @@
...
+    debug('item %d: sloppy converted %j to %j', ix, value[ix], itemResult);
+    if (itemResult.error)
+      return itemResult;
+    items[ix] = itemResult.value;
+  }
+  return { value: items };
+};
+
+ArrayConverter.prototype.validate = function(ctx, value) {
+  if (value === undefined || value === null)
```
Amir-61: ditto

lib/types/array.js
```diff
@@ -0,0 +1,157 @@
...
+function validateConverterResult(result) {
+  var isValid = typeof result === 'object' &&
+      ('error' in result || 'value' in result);
+  if (isValid)
+    return result;
+
+  var err = new (assert.AssertionError)({
+    message:
+      'Type conversion result should have "error" or "value" property. ' +
+      'Got ' + JSON.stringify(result) + ' instead.',
```
Amir-61: Question: Shouldn't it be globalized?

lib/context-base.js
bajtos: Yes, failed assertions dump the stack trace. That way it's easy to find the place in user code that's calling `ContextBase` with incorrect arguments.

lib/context-base.js
bajtos: There should be end-of-line character at the end of the file. GitHub UI is smart to not display it. If this character was missing, GitHub would display a red symbol.

lib/context-base.js
bajtos: Well, I want to make it clear in this file what are the method arguments `(name, value)`. In that sense, each empty method in `ContextBase` class is different, none of them is a generic NOOP.

lib/http-context.js
bajtos: No, assertion errors are not globalized.

lib/number-checks.js
bajtos: good point, will add.

lib/number-checks.js
bajtos: I think it doesn't really matter. MAX_SAFE_INTEGER is a well-known constant and as such it's ok to represent it as a static number value. No need to re-compute it all the time.

lib/context-base.js
bajtos: There shouldn't be an empty line (two end-of-line characters) at EOF.

lib/number-checks.js
bajtos: The EOL character is already there, see https://github.com/strongloop/strong-remoting/pull/343#discussion_r77783768

lib/shared-method.js
bajtos: Good catch. `debug` prints any extra arguments the same way as `console.log`, but then the timing info is inserted int the middle of the message. I'll fix this.

lib/shared-method.js
bajtos: The distinction between `uarg == null` and `uarg === null` is not immediately obvious and it's easy to miss when reading the code. I prefer to make this check explicit.

lib/type-registry.js
bajtos: I don't think we have a rule for that (yet). Since `eslint` checks are passing, the style should be good.

TBH, I don't really mind. If you think I should add curly braces, then I'll do so. Please let me know.

bajtos:
> I reviewed almost one thirds of your pull request tonight; I added a bunch of comments, mostly questions and some suggestions. I'll try to review the rest tomorrow to unblock you :-)

@Amir-61 you are awesome, thank you!

lib/types/any.js
bajtos: Yes, only lowercase `null` is recognized right now. Do you think this should be changed?

lib/types/array.js
bajtos:
```
$ node
> console.log('%s', new Error('asd'))
Error: asd
```

In this particular case, I think the stack trace is not that useful. If you think differently, then I can change the last argument to `err.stack || err`.

lib/types/array.js
bajtos: No, we don't globalize assertion errors.

lib/type-registry.js
```diff
@@ -0,0 +1,118 @@
...
+  this._options = options || {};
+  this._types = Object.create(null);
+  this._registerBuiltinTypes();
+}
+
+TypeRegistry.prototype.registerType = function(typeName, converter) {
+  assert(typeof typeName === 'string' && typeName,
+    'typeName must be a non-empty string');
+  assert(typeof converter === 'object' && converter,
+    'converter must be an object');
```
Amir-61: Are you just checking the type of `converter`. In this case shouldn't be like the following?
```
  assert(typeof converter === 'object',
     'converter must be an object');
```

lib/type-registry.js
Amir-61: Thats fine; TBH I like it too 👍  IIRC eslint used to be complaining about that.

lib/context-base.js
Amir-61: >Well, I want to make it clear in this file what are the method arguments (name, value). In that sense, each empty method in ContextBase class is different, none of them is a generic NOOP.

That makes sense; you may also consider a named function:
```
ContextBase.prototype.setReturnArgByName = function NOOP(name, value) {}
```
Please feel free to ignore it.

lib/number-checks.js
Amir-61: >No need to re-compute it all the time.
👍

lib/shared-method.js
Amir-61: >The distinction between uarg == null and uarg === null is not immediately obvious and it's easy to miss when reading the code. I prefer to make this check explicit.

I agree!

lib/types/array.js
Amir-61: Please ignore ^ based on your previous comment explaining your intention:
>The distinction between uarg == null and uarg === null is not immediately obvious and it's easy to miss when reading the code. I prefer to make this check explicit.

Amir-61:
Hey @bajtos,
I went through all changes and changes LGTM; I'm just not 100% sure if the changes break things in LoopBack core module which can be hard to detect.

lib/context-base.js
bajtos: When you assign a function to a class/prototype property, the stack trace use the property name as a function name.

```
$ node
> function ContextBase(){}
undefined
> ContextBase.prototype.foo = function(){ throw new Error(); }
[Function]
> new ContextBase().foo()
Error
    at ContextBase.foo (repl:1:47)
    at repl:1:19
    at sigintHandlersWrap (vm.js:22:35)
    at sigintHandlersWrap (vm.js:96:12)
    at ContextifyScript.Script.runInThisContext (vm.js:21:12)
    at REPLServer.defaultEval (repl.js:313:29)
    at bound (domain.js:280:14)
    at REPLServer.runBound [as eval] (domain.js:293:12)
    at REPLServer.<anonymous> (repl.js:504:10)
    at emitOne (events.js:101:20)
```

lib/type-registry.js
bajtos: Unfortunately, `typeof null` is also `object`:

```
> var converter = null
undefined
> typeof converter === 'object'
true
```

Amir-61: Oh right...I missed that! 👍

PR #344, opened 2016-09-05
[SEMVER-MAJOR] Fix Date serialization

bajtos:
Change the format of Date "returns" arguments from the output of `.toString()`, which produces values in local timezone, to the output of `.toJSON()`, which produces values in GMT.

Before:

    {
      "$type": "date",
      "$data": "Thu Jan 01 1970 01:00:00 GMT+0100 (CET)"
    }

Now:

    {
      "$type": "date",
      "$data": "1970-01-01T00:00:00.000Z"
    }

Connect to strongloop/loopback#1990

@ritch please review
@gunjpan @richardpringle @deepakrkris PTAL too

Note: in the issue strongloop/loopback#1990, I am proposing a more radical change. Unfortunately, it also requires more time which we don't have right now.

### DIFF
--- a/3.0-RELEASE-NOTES.md
+++ b/3.0-RELEASE-NOTES.md
@@ -176,3 +176,35 @@ correctly encodes request parameters.
 
 See [pull request #343](https://github.com/strongloop/strong-remoting/pull/343)
 for more details.
+
+## Serialization of Date values in responses
+
+The format of date values has been changed from the output of `.toString()`,
+which produces values in local timezone, to the output of `.toJSON()`, which
+produces values in GMT.
+
+For example, let's take `new Date(0)` returned for `dateArgument`.
+
+In strong-remoting 2.x, this value is converted to the following response
+when running on a computer in the Central-European timezone:
+
+```json
+{
+  "dateArgument": {
+    "$type": "date",
+    "$data": "Thu Jan 01 1970 01:00:00 GMT+0100 (CET)"
+  }
+}
+```
+
+In strong-remoting 3.x, the same value is converted to the following response
+regardless of the server timezone settings:
+
+```json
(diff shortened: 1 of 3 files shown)


ritch:
Are there complaints / issues with the existing support? This could break applications in a very subtle way that will be hard to detect.

That being said toJSON is the ideal format

bajtos:
> Are there complaints / issues with the existing support? 

Not really. We discovered this issue while working on iOS code generator and also while I was working on the rest-coercion test suite.

> This could break applications in a very subtle way that will be hard to detect.

TBH, I don't think there are many users relying on this feature. It requires a remote methods that has a "returns" argument with type "date" and there are no such methods in LoopBack. Users would have to write their own custom method.

I also believe this change is not as bad is it may look at the first sight. Both strings `Thu Jan 01 1970 01:00:00 GMT+0100 (CET)` and `1970-01-01T00:00:00.000Z` are a different representation of the same point in time.

I think there are only two things that can happen:
 - Clients capable of parsing both date formats (e.g. `new Date(str)` in JavaScript)` will not notice any difference.
 - Clients that understand the first format only should start failing (complaining about invalid date string).

It is also possible to have a client that silently ignores parsing errors and returns arbitrary date value in such case, but I would argue that's a bug in the client implementation.

@ritch What's your verdict? Merge or close?

richardpringle:
I think the change makes sense. @ritch also makes a good point that it could disrupt clients that aren't parsing the string "properly" but I don't think that we should cater to them.

I think that as long as the change is highlighted that it could be a breaking change in the release notes it should be fine.

ritch:
Merge it.

bajtos:
@slnode test please

PR #345, opened 2016-09-06
Fix tests to use res.get instead of res.headers

bajtos:
The latter is case-sensitive and expects header names in lower-case.

See also the discussion in #293 

@STRML please review


STRML:
:shipit:

PR #347, opened 2016-09-12
[SEMVER-MAJOR] Fix object type conversion

bajtos:
- Reject array/date values for object arguments
- Add rest-coercion tests for custom object types

This patch builds on top of recently landed #343

Connect to #312
Connect to strongloop-internal/scrum-loopback#974
Connect to strongloop-internal/scrum-loopback#885

@Amir-61 or @richardpringle please review
cc @ritch @STRML

### DIFF
--- a/lib/types/any.js
+++ b/lib/types/any.js
@@ -8,6 +8,7 @@
 var debug = require('debug')('strong-remoting:http-coercion');
 var g = require('strong-globalize')();
 var numberChecks = require('../number-checks');
+var looksLikeJson = require('../looks-like-json').looksLikeJson;
 
 var MAX_SAFE_INTEGER = numberChecks.MAX_SAFE_INTEGER;
 var MIN_SAFE_INTEGER = numberChecks.MIN_SAFE_INTEGER;
@@ -46,12 +47,18 @@ module.exports = {
         value = num;
     }
 
-    var objectConverter = ctx.typeRegistry.getConverter('object');
-    var objectResult = objectConverter.fromSloppyValue(ctx, value);
-    if (!objectResult.error && objectResult.value)
-      return objectResult;
+    if (looksLikeJson(value)) {
+      try {
+        var result = JSON.parse(value);
+        debug('parsed %j as JSON: %j', value, result);
+        return this.fromTypedValue(ctx, result);
+      } catch (ex) {
+        debug('Cannot parse "any" value %j, assuming string. %s', value, ex);
+        // no-op, use the original string value
+      }
+    }
 
-    return { value: value };
+    return this.fromTypedValue(ctx, value);
   },
 
   validate: function(ctx, value) {
--- a/3.0-RELEASE-NOTES.md
+++ b/3.0-RELEASE-NOTES.md
@@ -173,6 +173,7 @@ Most notable breaking changes:
    the value as a number now. That way the value "0" always produces
    "1970-01-01T00:00:00.000Z" instead of some date around 1999/2000/2001
    depending on server timezone.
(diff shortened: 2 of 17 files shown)


lib/types/any.js
```diff
@@ -46,12 +46,22 @@ module.exports = {
         value = num;
     }
 
-    var objectConverter = ctx.typeRegistry.getConverter('object');
-    var objectResult = objectConverter.fromSloppyValue(ctx, value);
-    if (!objectResult.error && objectResult.value)
-      return objectResult;
+    var looksLikeJson = typeof value === 'string' && (
```
STRML: This seems like it could be a nice utility function to pull out.

STRML:
LGTM, I agree that while JS says `new Date() instanceof Object` and `Array() instanceof Object`, that's nonsensical and clearly not what any developer would actually want.

lib/types/any.js
bajtos: Agreed, will extract.

richardpringle:
I completely agree for dates, but some might be using any JSON notation as type-object, which of course includes an array. 

I personally don't think that we should be so strict, if a developer wants to assure that an object strictly starts with `{`, they should implement the type check themselves.

bajtos:
> I completely agree for dates, but some might be using any JSON notation as type-object, which of course includes an array. 

One can use `type: 'any'` for arguments that support multiple types.

> I personally don't think that we should be so strict, if a developer wants to assure that an object strictly starts with `{`, they should implement the type check themselves.

My issue with this is that most developers won't realise their function can receive an array in an object argument.

IMO, the case where an array is not wanted will be much more common than the case where a method expects either an object or an array.

I am proposing to implement a new flag that will signal that an argument can accept both types and defer the implementation until there is a LoopBack user asking for this feature.

```js
{ arg: 'data', type: 'object', allowArrays: true }
```

@richardpringle Thoughts?
cc @STRML @ritch

bajtos:
> My issue with this is that most developers won't realise their function can receive an array in an object argument.

For example, let's take a LoopBack model called `Product`:

```js
$ node
> new Product([{ name: 'Pen' }, { name: 'Pencil' }])
{ '0': { name: 'Pen' }, '1': { name: 'Pencil' } }
```

Because the model implementation assumes the input is an object, it treats the array indices as property names and array items as property values. I don't think that's a result the caller was expecting!

richardpringle:
I like the flag, because that allows anything in JSON format

STRML:
Why not unions? Seems better to allow ['object', 'array'] than a special
flag.

bajtos:
> Why not unions? Seems better to allow ['object', 'array'] than a special flag.

Even better 👍 

(Note that Swagger spec does not support union types yet, AFAIK.)

I still think this should be kept out of scope of this pull request though.

STRML:
Yeah I agree.

In the future I'd like to see the ability to provide functions as types - could add the following extensions mirroring Model validators:

```js
// First arg must match a provided arg in the remoting function
User.update.validate('firstname', function(err) {...})
```

And possibly even something to override coercion?

```js
User.update.coerce('firstname', function(value) {...})
```

This way we could maintain swagger compat by still requiring a string type name.

Amir-61:
>Why not unions? Seems better to allow ['object', 'array'] than a special
flag.

Thats a great idea 👍  I agree we should keep it out of the scope of this PR.

LGTM

richardpringle:
Might be a good idea to accept any valid JSON with type-JSON too (also beyond the scope of the PR).

bajtos:
Thank you all for your input.

Issue #238, opened 2015-08-27
Timestamps not converted to Date anymore


kflip:
Timestamps passed as parameter to remote methods worked in older versions:
e.g.: 
	from=1440631417
or in milliseconds
	from=1440631417000

It fails to instantiate a new Date() due a missing conversion from String to Number.
I've made a workaround in Shared-Method.js, but you might want to solve this somewhere else (HttpContext) or differently.

thanks
	
	...
	function convertValueToTargetType
	...
	case 'date': {
		
		// could be a timestamp as string
		if (!isNaN(value) && typeof value == 'string') {
		    value = Number(value).valueOf();
 	    } 
	
		return new Date(value);
	}

richardpringle:
@bajtos, can you confirm if this is expected behavior or not?

If we don't expect a number to coerce, then I think there should be a warning in the docs.

bajtos:
Interesting. So if my understanding is correct, when we have `{ accepts: { arg: 'd', type: 'Date' } }`,
it used to be possible to send `?from=1440631417`, where the value was treated as timestamp. @kflip could you please confirm? Do you happen to know the last version where this was still working?

bajtos:
This is actually tricky. Both `new Date('0')` and `new Date(0)` are valid ways how to create a new Date instance, however each returns a different value. It makes sense to me to always convert timestamp-like values to numbers before creating a Date value.

I have a question though. How to handle numeric values in scientific notation (e.g. `?ts=1.234e%2B30`) and values that cannot be accurately represented as integer numbers (e.g. `?ts=2343546576878989879789`)? I am proposing to reject such values and return HTTP 400 Bad Request response to the client.

/cc @STRML

richardpringle:
@bajtos can this be closed now since #349 was landed?

bajtos:
Yes, I am closing this issue as fixed (in LoopBack 3.0).

PR #349, opened 2016-09-13
[SEMVER-MAJOR] Fix rest-coercion of timestamp strings

bajtos:
When converting a value coming from a string source like querystring to a date, and the value is a timestamp (i.e. an integer), we treat the value as a number now.

That way the value "0" always produces "1970-01-01T00:00:00.000Z" instead of some date around 1999/2000/2001 depending on server timezone.

Connect to #238

@Amir-61 @richardpringle please review
cc @STRML @ritch


richardpringle:
LGTM

Issue #267, opened 2015-12-15
Application/json body request with an extended content-type header (charset) should not be coerced


ebourmalo:
The initial check to turn off the coercion of the request body when the header is `application/json` was:
```javascript
ctx.req.get('content-type') === 'application/json'
```

If the `content-type` header is extended with the charset definition (i.e `application/json;charset=UTF-8`), the body is coerced and that is NOT what we want :)

Here is the PR for a fix, updating the check for allowing this use case: https://github.com/strongloop/strong-remoting/pull/266

STRML:
This is also fixed in #264.

PR #350, opened 2016-09-13
Detect JSON content-type with parameters

bajtos:
Fix the detection of JSON content type used to decide whether sloppy
coercion should be applied, so that it correctly supports values with
parameters. For example:

    application/json;charset=UTF-8

Connect to #267

@richardpringle please review
cc @ebourmalo


richardpringle:
LGTM

PR #352, opened 2016-09-14
[SEMVER-MAJOR] Remove built-in CORS middleware

bajtos:
Following the discussion in https://github.com/strongloop/strong-remoting/pull/293, I would like to revisit the way how CORS is configured in LoopBack applications.

There are two types of applications that are commonly built using LoopBack, and each type has different requirements when it comes to CORS.

 1. Public facing API servers serving a variety of clients. CORS should be enabled to allow 3rd party sites to access the API server living on a different domain than the client(s).
 2. API+[SPA](https://en.wikipedia.org/wiki/Single-page_application) sites where both the API server and the SPA client live on the same domain and the API should be locked down to serve only its own browser clients, i.e. CORS must be disabled.

So far, we were enabling CORS by default to make it easy to build public facing API servers. Unfortunately, this was also making API+SPA sites powered by strong-remoting unsecure by default. What's even worse, this insecure setting was not immediately obvious.

**In this pull request, I am removing CORS from strong-remoting, to push the responsibility of enabling/configuring CORS back to the application developer.**

Note that this change does not affect LoopBack applications scaffolded by a recent version of `slc loopback` or `apic loopback`, as these applications don't rely on the built-in CORS middleware - they configure CORS explicitly in `server/middleware.json`. (See the project template in loopback-workspace, it is disabling the built-in CORS middleware [here](https://github.com/strongloop/loopback-workspace/blob/3e80fcd3053199c5de549a5ded5e7773ad2fc5cc/templates/projects/empty-server/data.js#L82) and explicitly configuring CORS for the entire server [here](https://github.com/strongloop/loopback-workspace/blob/3e80fcd3053199c5de549a5ded5e7773ad2fc5cc/templates/projects/empty-server/files/server/middleware.json#L7-L13).

@richardpringle @STRML please review
cc @gunjpan @deepakrkris @raymondfeng @ritch

### DIFF
--- a/test/rest.test.js
+++ b/test/rest.test.js
@@ -299,7 +299,7 @@ describe('strong-remoting-rest', function() {
     });
   });
 
-  describe('cors', function() {
+  describe('CORS', function() {
     var method;
     beforeEach(function() {
       method = givenSharedStaticMethod(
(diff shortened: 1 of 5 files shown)


test/rest.test.js
```diff
@@ -319,89 +319,34 @@ describe('strong-remoting-rest', function() {
...
-      var url = server.url.replace('/testClass/testMethod', '');
-      server
-        .set('Accept', 'application/json')
-        .set('Content-Type', 'application/json')
-        .set('Origin', url)
-        .send({ person: 'ABC' })
-        .end(function(err, res) {
-          assert(res.get('Access-Control-Allow-Origin') === undefined);
-          assert(res.get('Access-Control-Allow-Credentials') === undefined);
+        .expect(200, function(err, res) {
```
richardpringle: When you make a cross origin request, you wouldn't get a 200 would you?

richardpringle:
Was the affect on API Connect taken into consideration? I know that CORS support is configured at the gateway level there, just something to consider if it hasn't been already.

test/rest.test.js
deepakrkris: 200 may not be the right http status code, test should check for the appropriate error code

bajtos:
Thank you @richardpringle and @deepakrkris for the review.

> Was the affect on API Connect taken into consideration? I know that CORS support is configured at the gateway level there, just something to consider if it hasn't been already.

Yes. There are two parts to the integration with API Connect.

As I explained in the pull request description, applications scaffolded by our generator (`apic loopback`) setup their own CORS middleware for the entire server. As a result, they are not affected by this change.

When the LoopBack app is run behind APIC gateway, the gateway will provide CORS headers as specified in Gateway's config. As a result, any application-provided CORS headers should be ignored.

I was pondering the idea of disabling CORS in the LoopBack application that are expected to run in APIC environment, but @raymondfeng mentioned good reasons why we should keep CORS in the LoopBack app too:

> 1. APIC or other tools might serve API explorer that talks to LoopBack app directly.
> 2. Internal web apps might interact with LoopBack app without going through the gateway.

test/rest.test.js
bajtos: In my understanding of CORS, it's the responsibility of the browser (the client) to enforce cross-origin policies. The server is required to describe which origins are allowed to access it from the browser.

See https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS (emphasis is mine)
> The Cross-Origin Resource Sharing standard works by adding new HTTP headers that **allow servers to describe the set of origins that are permitted to read that information using a web browser**.  Additionally, for HTTP request methods that can cause side-effects on user data (in particular, for HTTP methods other than GET, or for POST usage with certain MIME types), the specification mandates that browsers "preflight" the request, soliciting supported methods from the server with an HTTP OPTIONS request method, and then, upon "approval" from the server, sending the actual request with the actual HTTP request method.  Servers can also notify clients whether "credentials" (including Cookies and HTTP Authentication data) should be sent with requests.

In our tests here, we are not acting like a browser (the `supertest` library is an HTTP agent only, not a browser-like client) and therefore the status code `200` is expected.

bajtos:
> Review required
> At least one approved review is required.

I have enabled GitHub checks to require each pull request to get an approval before it can be merged. @richardpringle @deepakrkris when the code LGTY, please use the new pull request UI ([docs](https://help.github.com/articles/about-pull-request-reviews/)) for approving the patch.

bajtos:
The build is failing on `windows-0.12` because of a problem on the CI, I am going to ignore the failure and land the pull request.

deepakrkris:
LGTM

PR #355, opened 2016-09-20
Don't convert arg values returned by http function

bajtos:
When an argument provides "http" mapping as a custom function, then we should not traverse the return value to detect buffers and dates. The return value could be a complex object, e.g. the full remoting context, in which case the traversal takes too long (in order of seconds!).

I discovered this issue while playing with a solution for injecting context via `options` argument (see https://github.com/strongloop/loopback/issues/1495 and the pull request https://github.com/strongloop/loopback/pull/2762).

@ritch @richardpringle or @deepakrkris please review


ritch:
👍

PR #357, opened 2016-09-21
Deprecate built-in CORS middleware

bajtos:
Push the responsibility of enabling/configuring CORS back to the application developer.

Back-port #352

@richardpringle or @deepakrkris please review

### DIFF
--- a/test/rest.browser.test.js
+++ b/test/rest.browser.test.js
@@ -30,7 +30,7 @@ describe('strong-remoting-rest', function() {
 
   // setup
   beforeEach(function() {
-    objects = RemoteObjects.create();
+    objects = RemoteObjects.create({cors: false});
     remotes = objects.exports;
 
     // connect to the app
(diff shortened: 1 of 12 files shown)


test/rest.browser.test.js
```diff
@@ -30,7 +30,7 @@ describe('strong-remoting-rest', function() {
 
   // setup
   beforeEach(function() {
-    objects = RemoteObjects.create();
+    objects = RemoteObjects.create({cors: false});
```
richardpringle: Is it necessary to set CORS in all the tests? Should these tests not use the default settings?

richardpringle:
The deprecation messages look fine

test/rest.browser.test.js
bajtos: The default setting is to enable the built-in CORS middleware with `{ origin: true, credentials: true }`, which triggers deprecation warnings. That was my intention, to poke 2.x users to rework the CORS settings in their application soon.

This applies to our tests too. If I didn't add `{cors: false}` to all the tests, then `npm test` prints a lot of deprecation warnings.

Considering the amount of changes in https://github.com/strongloop/loopback-component-explorer/pull/178 too, perhaps it was not a good idea to add this deprecation message?

richardpringle: Ah! I didn't think about those messages crowding the test results!

I think you should leave them in.

richardpringle:
approved!

PR #359, opened 2016-09-22
Add translation files

Amir-61:
* Add translation files for 2.x


.gitignore
```diff
@@ -11,5 +11,3 @@
 node_modules/
 /coverage/
 dist
```
0candy: Is there an EOL here?

Amir-61: Yes there was one. I changed it to two and now GH shows it lol

0candy:
LGTM

PR #363, opened 2016-09-30
Fix support for hooks returning a Promise

bajtos:
Fix the code to correctly recurse and invoke next hooks after the promise returned by the first hook callback was resolved.

The issue was discovered by @tvdstaaij in https://github.com/strongloop/loopback/pull/2751

@gunjpan @richardpringle or @deepakrkris please review


bajtos:
Before the fix, the modified unit-test was failing with the following error:

```
  1) strong-remoting-rest call of prototype method should resolve promise returned by a hook:

      Uncaught AssertionError: expected [ 'first' ] to deeply equal [ 'first', 'second' ]
      + expected - actual

       [
         "first"
      +  "second"
       ]
```

gunjpan:
LGTM

Issue #360, opened 2016-09-23
Allow both object and array values for an argument (create/createMany)


davidcheung:
continuing the conversation for this comment https://github.com/strongloop/strong-remoting/pull/347#issuecomment-246680808

Since #347, strong-remoting rejects array values for arguments of type object. As a result, it's no longer possible to send an array of objects via `POST /api/my-model-/create` and expect the server to create them all in a batch mode.

### Background

with https://github.com/strongloop/strong-remoting/blob/master/3.0-RELEASE-NOTES.md#conversion-and-coercion-of-input-arguments
*Array values are not allowed for arguments of type object.* 
https://github.com/strongloop/loopback/blob/ed8b651b988799b5b42289bddd9f568398776bc8/lib/persisted-model.js#L643

### Why its a problem

I think the biggest breaking change brought from this would be the ability to do batch `POST` operations

### Possible options

- Option 1
https://github.com/strongloop/strong-remoting/pull/347#issuecomment-246680808
add flag to `allowArrays` for object coercion 
__i would go with this as it's least painful for migration__

- Option 2
change type of persistetedModel methods that supports batch operations to 'any'

- Option 3 
Create `POST /createMany` and take `array of objects`

@strongloop/loopback-dev

bajtos:
Since #347, strong-remoting rejects array values for arguments of type object. As a result, it's no longer possible to send an array of objects via LoopBack's built-in `POST /api/my-model-/create` and expect the server to create them all in a batch mode.

> Option 1
> add flag to allowArrays for object coercion 

I think the discussion in #347 reached an agreement that a union type ("object" or "array") is better than "allowArrays" flag. 

In that light, I am proposing to go with Option 3, as it provides API that's easier to describe in Swagger: **Create `POST /createMany` and take `array of objects`**

As for Option 2, *change type of persistetedModel methods that supports batch operations to 'any'*, I am afraid this would break the swagger description completely.

@richardpringle @Amir-61 @STRML any opinions?

richardpringle:
Type JSON! 

There should be a type that accepts any valid JSON which is pretty much the same as saying "array or object"!

I do not thing that more endpoints is the answer, especially just to deal with strong-remoting types... the function is the exact same yet we have two endpoints because we have to allow two different input types. 

I think that the default Persisted model REST API should be simple and minimalist.  Adding a new endpoint to solve a problem should be an absolute last resort.

richardpringle:
[...]

PR #369, opened 2016-10-11
allowArray flag for remoteObject `accept`

davidcheung:
fixes https://github.com/strongloop/strong-remoting/issues/360

In loopback 2.x endpoints supported batch create which passes
an Array of Objects into the endpoint that expects Objects,
this is nolonger allowed in 3.x's stricter coercion
In order to maintain backwards compatiability, this is a short term
fix to allow endpoints to accept Array of Objects with the allowArray

### DIFF
--- a/lib/shared-method.js
+++ b/lib/shared-method.js
@@ -225,9 +225,10 @@ SharedMethod.prototype.invoke = function(scope, args, remotingOptions, ctx, cb) 
       var desc = accepts[i];
       var name = desc.name || desc.arg;
       var uarg = SharedMethod.convertArg(desc, args[name]);
+      var conversionOptions = SharedMethod.getConversionOptionsForArg(desc);
 
       try {
-        uarg = validateInputArgument(uarg, desc, ctx);
+        uarg = validateInputArgument(uarg, desc, ctx, conversionOptions);
       } catch (e) {
         debug('- %s - ' + e.message, sharedMethod.name);
         return cb(e);
@@ -306,7 +307,7 @@ function escapeRegex(d) {
  * @param  {Context} ctx       Remoting request context.
  * @return {*}                 Coerced argument.
  */
-function validateInputArgument(uarg, desc, ctx) {
+function validateInputArgument(uarg, desc, ctx, conversionOptions) {
   var name = desc.name || desc.arg;
 
   // Verify that a required argument has a value
@@ -318,7 +319,7 @@ function validateInputArgument(uarg, desc, ctx) {
   }
 
   var converter = ctx.typeRegistry.getConverter(desc.type);
-  var err = converter.validate(ctx, uarg);
+  var err = converter.validate(ctx, uarg, conversionOptions);
   if (err) {
     err.message = g.f('Invalid argument %j. ', name) + err.message;
     throw err;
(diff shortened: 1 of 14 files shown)


davidcheung:
@bajtos PTAL,  this is so hacky 😭  
any thoughts on how to improve the implementation?

bajtos:
> PTAL, this is so hacky

Yeah, I agree.

How about adding an `options` object to all three converter methods (`fromSloppyValue`, `fromTypedValue`, `validate`)? It will require one-time change in all converters to ensure `options` are passed down to other methods (please send it in a standalone commit), but then the implementation of `allowArray` should be very simple.

Thoughts?

davidcheung:
@slnode test please

davidcheung:
@bajtos PTAL again, i think this is better this time

lib/types/object.js
```diff
@@ -50,8 +51,25 @@ module.exports = {
...
+        return errorNotAnObject();
+      } else {
+        var err = null;
+        value.forEach(function(item) {
+          if (typeof item !== 'object' || Array.isArray(item) ||
+          item === null) {
+            err = errorArrayItemsNotAnObject();
+            return;
+          }
+        });
```
bajtos: I think this may be a bit cleaner:

```js
var hasInvalidItems = value.some(function(item) {
  return typeof item !== 'object' || Array.isArray(item) || item === null;
});

return hasInvalidItems ? errorArrayItemsNotAnObject() : null;
```

How about Dates as the array items, I think we should reject them too. Can we perhaps reuse `validate` inside the forEach/some callback?

```js
var self = this;
var hasInvalidItems = value.some(function(item) {
  return !self.validate(item/* no options (?) */);
});
return hasInvalidItems ? errorArrayItemsNotAnObject() : null;
```

lib/types/object.js
```diff
@@ -65,3 +83,9 @@ function errorNotAnObject() {
   err.statusCode = 400;
   return err;
 }
+
+function errorArrayItemsNotAnObject() {
+  var err = new Error(g.f('Array items are not an object.'));
```
bajtos: How about `Some of array items are not an object` ?

test/rest-coercion/jsonbody-object.suite.js
```diff
@@ -75,4 +75,20 @@ module.exports = function(ctx) {
       [[1, 2], ERROR_BAD_REQUEST],
     ]);
   });
+
+  describe('json body - object - allowArray: true', function() {
+    verifyTestCases({ arg: 'data', type: 'object', allowArray: true }, [
+      // should accept regular objects, plus array of objects
```
bajtos: could you please add few cases for "regular objects"?

davidcheung:
oh nice, changed now auto collapse again, @bajtos PTAL
applied the feedback and also added test case in json-form and customClass

lib/shared-method.js
```diff
@@ -636,3 +637,20 @@ SharedMethod.prototype.addAlias = function(alias) {
     this.aliases.push(alias);
   }
 };
+
+/**
+ * build conversion options from remote's' args
+ *
+ * @param arg (remote args such as `accepts`)
+ * @param optionsObj (optional parameter that gets operated on if you provide it)
```
bajtos: please remove

lib/shared-method.js
```diff
@@ -636,3 +637,20 @@ SharedMethod.prototype.addAlias = function(alias) {
     this.aliases.push(alias);
   }
 };
+
+/**
+ * build conversion options from remote's' args
+ *
+ * @param arg (remote args such as `accepts`)
```
bajtos:
```
@param {Object} arg Definition of accepts/returns argument.
```

lib/shared-method.js
```diff
@@ -636,3 +637,20 @@ SharedMethod.prototype.addAlias = function(alias) {
...
+ * @param optionsObj (optional parameter that gets operated on if you provide it)
+ * @returns optionsObj(options to pass to typeConverter methods, eg: validate, fromTypedValue)
```
bajtos:
```
@returns {Object} Options object to pass to type-converter methods,
   e.g `validate` or `fromTypedValue`.
```

davidcheung:
ci failing due to `http-auth` dropping node 10 / 12 support :sob: https://github.com/http-auth/http-auth/commit/0097ed6195986942a65e14b38b3f30281c9f6435

davidcheung:
@slnode test please

davidcheung:
@slnode test please

davidcheung:
@slnode test please

davidcheung:
@slnode test please

Issue #318, opened 2016-07-07
Add support for coercion by endpoints with type File


devInterests:
If you have a remote method with a parameter that's type: 'file' you receive following error:
`"Error: Invalid value for argument 'file' of type 'file': [object Object]. Received type was converted to object.`

As far as I understood the error is in function convertToBasicRemotingType in shared-method.js:

```
function convertToBasicRemotingType(type) {
  if (Array.isArray(type)) {
    return type.map(convertToBasicRemotingType);
  }
  if (typeof type === 'object') {
    type = type.modelName || type.name;
  }

  type = String(type).toLowerCase();

  switch (type) {
    case 'string':
    case 'number':
    case 'date':
    case 'boolean':
    case 'buffer':
    case 'object':
    case 'file':
    case 'any':
      return type;
    case 'array':
      return ['any'].map(convertToBasicRemotingType);
    default:
      // custom types like MyModel
      return 'object';
  }
}
```

The type `file` is supported as a separate type. The problem is that the arg is coming as type `object`.
My fix at the moment is to use `object` as arg type or to use a older version where the file type is not supported and becomes default type `object`.
But i would like to use the correct type: `file`.

My interface definition:

```
model.remoteMethod( 
       'fileUpload',
        {
            http:{ verb:'Post', source: 'body', path: '/upload/:id'},
            description: '',
            accepts: [
                {arg: 'id', type: 'string', http: { source: 'path' }, required: true},
                {arg: 'file', type: 'file', http: { source: 'body' }, required: true},
                {arg: 'req', type: 'object', http: {source: 'req'}}
            ],
            returns: {arg: 'res', type: 'object', root:true}
        });
```

Perhaps I am doing something wrong?

richardpringle:
Hey @devInterests sorry about the delayed response.

The `accepts.type` must a [LoopBack type](https://docs.strongloop.com/display/public/LB/LoopBack+types). 

I am going to close this issue for now but please feel free to carry on the conversation in here or re-open if you believe this is still an issue.

devInterests:
Hello,

@richardpringle I was aware of the list.
It seems I misunderstood the answer: [Uploading images](https://groups.google.com/forum/#!topic/loopbackjs/-7xumo0V8qg)

> 1. We allow remote method parameters to be typed as ‘File’. If one of the parameters is typed as ‘File’, only ‘multipart/form-data’ content type is supported.

It sounds to me like the File-Type is supported.

[...]

PR #370, opened 2016-10-13
Distinguish between "file" and "File" result types

fabien:
If you have a Loopback model called `File` you'll receive following error when it is set as a return type: `Error: Cannot create a file response from object`

In other words, because `file` is now a reserved keyword with a very specific response handling logic, and `File` is lowercased during normalization into `resultType`, the error above would prevent any Loopback being named `File`. This introduced a breaking change for my apps.

This is loosely related to #318 but in the above case, it's actually not desirable to return a file (buffer/stream) response. Perhaps a more specific keyword than simply `file` would have been better in hindsight.

See also: https://github.com/strongloop/strong-remoting/commit/a85068ef1122751efbbf3e9b2a1b9439ae70fa3f#commitcomment-19410422

@bajtos @richardpringle

### DIFF
--- a/lib/http-context.js
+++ b/lib/http-context.js
@@ -412,9 +412,7 @@ HttpContext.prototype.setReturnArgByName = function(name, value) {
   }
 
   if (returnDesc.root) {
-    // TODO(bajtos) call SharedMethod's convertToBasicRemotingType here?
-    this.resultType = typeof returnDesc.type === 'string' ?
-      returnDesc.type.toLowerCase() : returnDesc.type;
+    this.returnsFile = SharedMethod.isFileType(returnDesc.type);
     return;
   }
 
@@ -651,7 +649,7 @@ HttpContext.prototype.done = function(cb) {
     res.header('Content-Type', operationResults.contentType);
   }
   if (dataExists) {
-    if (this.resultType !== 'file') {
+    if (!this.returnsFile) {
       operationResults.sendBody(res, data, method);
       res.end();
     } else if (Buffer.isBuffer(data) || typeof(data) === 'string') {
--- a/lib/shared-method.js
+++ b/lib/shared-method.js
@@ -404,6 +404,10 @@ function convertToBasicRemotingType(type) {
     type = type.modelName || type.name;
   }
 
+  if (SharedMethod.isFileType(type)) {
+    return 'file';
+  }
+
   type = String(type).toLowerCase();
 
   switch (type) {
@@ -414,7 +418,6 @@ function convertToBasicRemotingType(type) {
     case 'boolean':
     case 'buffer':
     case 'object':
-    case 'file':
     case 'any':
       return type;
     case 'array':
@@ -541,6 +544,12 @@ SharedMethod.convertArg = function(accept, raw) {
   return data;
 };
 
+SharedMethod.isFileType = function(argSpec) {
+  // Only all-lowercase value "file" is recognized as the special file type
+  // Other spellings, most notably "File", are treated as custom types (models)
+  return argSpec === 'file';
+};
+
 /**
  * Returns a reformatted Object valid for consumption as JSON from an Array of
  * results from a remoting function, based on `returns`.
@@ -564,7 +573,7 @@ SharedMethod.toResult = function(returns, raw, ctx) {
 
     if (item.root) {
       var targetType = convertToBasicRemotingType(item.type);
-      var isFile =  targetType === 'file';
+      var isFile = SharedMethod.isFileType(item.type);
       result = isFile ? raw[index] : convert(raw[index], targetType, item.name);
       return false;
     }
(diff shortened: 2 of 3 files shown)


fabien:
Related (more or less duplicate of): https://github.com/strongloop/loopback/issues/2554

bajtos:
@fabien thank you for the pull request. I am ok to land this patch as short-term fix.

For longer term: I think the situation where `type: 'file'` is something else than `type: 'File'` is quite confusing. Let's discuss how can we improve the format of remoting metadata in https://github.com/strongloop/loopback/issues/2554

fabien:
@bajtos thanks for the quick reply and aim to land it ASAP! With `strong-remoting` being a dependency of `loopback` a short-term fix for me would have involved forking both in order to fix the package.json referencing my fork on Github ... much appreciated. 

Will look into the discussion some more.

bajtos:
@fabien there are few more places where we are dealing with `type: 'file'`, see e.g. https://github.com/strongloop/strong-remoting/blob/1efefa62d5e4c393ffe631e560e1946f688acb24/lib/shared-method.js#L417, https://github.com/strongloop/strong-remoting/blob/1efefa62d5e4c393ffe631e560e1946f688acb24/lib/shared-method.js#L567 and https://github.com/strongloop/strong-remoting/blob/1efefa62d5e4c393ffe631e560e1946f688acb24/lib/shared-method.js#L578. PTAL.

lib/http-context.js
```diff
@@ -415,6 +415,7 @@ HttpContext.prototype.setReturnArgByName = function(name, value) {
     // TODO(bajtos) call SharedMethod's convertToBasicRemotingType here?
     this.resultType = typeof returnDesc.type === 'string' ?
       returnDesc.type.toLowerCase() : returnDesc.type;
+    this.isFile = returnDesc.type === 'file';
```
bajtos: Since `this` is the context, the name`isFile` is misleading in my opinion. How about `isReturningFile`?

fabien:
@bajtos I'll look into it

fabien:
@bajtos as you noted here: 

https://github.com/strongloop/strong-remoting/blob/2.x/lib/http-context.js#L415

Why aren't we using `convertToBasicRemotingType`?

fabien:
@bajtos I have gone ahead and use `convertToBasicRemotingType` and then prefix/normalize actual `file` types as `res:file` (meaning *reserved* or *response*). We can use a different prefix if you want. With this change, handling such custom types is now much more generic. If you wanted to explicit in your remote method declaration, you could add the prefix yourself:

```{ type: 'res:file' }```

And it would still work as expected. This means it's backwards compatible, because `file`, `File` and `res:file` will all be handled appropriately.

File: lib/http-context.js (hunk 1 above)
Comment (bajtos) [inline lib/http-context.js]: sure, that's fine too

File: lib/shared-method.js [hunk 2]
```diff
@@ -404,19 +404,24 @@ function convertToBasicRemotingType(type) {
...
     case 'date':
     case 'boolean':
     case 'buffer':
     case 'object':
-    case 'file':
     case 'any':
-      return type;
+      return normalizedType;
+    case 'file':
+      return type === 'File' ? normalizedType : ('res:' + normalizedType);
```
bajtos: IMO we need to use a different condition here: only `file` should be treated as a built-in file type, everything else is a custom model (`FILE`, FilE`, etc.).

bajtos:
@fabien I have a different opinion here. I would rather not add a new set of type strings like `res:file`, I am worried it will create even more confusion (why there is `object` and not `res:object` for example?).

As I commented in https://github.com/strongloop/loopback/issues/2554#issuecomment-253786224, I think we should simply make `File` a reserved type starting from 3.0 release, in which case this fix will stay in the `2.x` branch only.

Anyhow, I went ahead and simplified/cleaned up your patch in 7691ff7 to a state that looks good to myself. Could you please take a look at the resulting change? (I pushed the commit to your feature branch thanks to a recently added GitHub feature described in https://github.com/blog/2247-improving-collaboration-with-forks).

If you are ok with what we have here now, then either me or you can squash the commits into a single one, provide a descriptive commit message per our [contributor docs](http://loopback.io/doc/en/contrib/git-commit-messages.html), and this pull request can be landed.

fabien:
@bajtos thanks for the cleaned-up patch - I'm totally fine with your implementation! If you don't mind, please go ahead and squash & pull to get it landed 👍

fabien:
Only concern would be the 3.0 release upgrade path for models named `File`, but I guess that's unavoidable.

bajtos:
> Only concern would be the 3.0 release upgrade path for models named File, but I guess that's unavoidable.

Yes, I do realise this means users will have to rework their apps a bit when upgrading from 2.x to 3.0. As you said, this is unavoidable :(

bajtos:
Landed and released in `strong-remoting@2.32.2`, thank you for the contribution!

fabien:
@bajtos thanks for landing it so quickly!

PR #372, opened 2016-10-17
Stop changing read-only Number property

kobaska:
Trying to set value on a read only variable fails on older web views with the following error:

```
["Error group: EXCEPTION: Error: Uncaught (in promise): TypeError: Cannot assign to read only property 'MAX_SAFE_INTEGER' of function Number() { [native code] }"]
```

I have added some guard to make sure it's only set if undefined. This issue was noticed while running on Android 5.0 web views. Works on Android 6.

### DIFF
--- a/lib/shared-method.js
+++ b/lib/shared-method.js
@@ -732,14 +732,21 @@ SharedMethod.prototype.addAlias = function(alias) {
 };
 
 // To support Node v.0.10.x
-Number.isInteger = Number.isInteger || function(value) {
-  return typeof value === 'number' &&
-    isFinite(value) &&
-    Math.floor(value) === value;
-};
+if (typeof Number.isInteger === 'undefined') {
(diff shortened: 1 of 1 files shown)


slnode:

slnode:

slnode:

slnode:

bajtos:
@kobaska thank you for the pull request. I improved it a bit for you - reworked other `Number` shims too and improved the commit message to be more descriptive.

Could you PTAL at the result and let me know if it still looks good to you now? https://github.com/strongloop/strong-remoting/pull/372/files

FWIW, the problem is not present in 3.x, because we are not modifying the global `Number` property any more.

bajtos:
@slnode ok to test

bajtos:
Landed, thank you @kobaska for the contribution.

I have released the fix in `strong-remoting@2.32.3`

PR #374, opened 2016-10-18
lockdown dependency for http-auth

davidcheung:
Due to a recent release of 2.5.10 which broke node 10/12 support
this commit locks down dependency before 2.5.11 is released
as some registry will still pick up unpublished version (2.5.10)

See http-auth/http-auth@0097ed6195986942a65e14b38b3f30281c9f6435

----
edit found out this is causing node 10/12 to fail too
http-auth/apache-md5@4f3b24d3d382851ab2b72a4132f73a8311473856 😢

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "grunt-contrib-watch": "^1.0.0",
     "grunt-eslint": "^18.0.0",
     "grunt-karma": "^2.0.0",
-    "http-auth": "^2.2.5",
+    "http-auth": "2.4.11",
     "karma": "^1.2.0",
     "karma-browserify": "^5.1.0",
     "karma-chrome-launcher": "^1.0.1",


davidcheung:
2.4.11 was released with `apache-md5` and `apache-crypt` locked down as well,
looks like until we can drop support for node10/12 we will have to use `2.4.11`
https://github.com/http-auth/http-auth/commit/f318a7f2326615347dba0192c97265833f6644d3

cc @bajtos @rmg 

related https://github.com/strongloop/loopback/issues/2807

sam-github:
@davidcheung I don't see a link to the bug report of the semver violation in http-auth's issue tracker

davidcheung:
@sam-github i started the convo in the commit, then the version that violated the semver was already unpublished, so i didnt end up creating a bug report at all.

sam-github:
I'm confused, if the problem was fixed upstream by unpublishing the breaking change... why do you still need to lock down the dep spec here?

davidcheung:
@sam-github 
it was originally at 2.4.10 prior to anything breaking, 
then `2.5.10` was published (breaking node10/12) then unpublished
and `2.4.11` was published with the deps locked down from upstream, but caching proxies are still picking up `2.5.10`

sam-github:
got it, thanks.

PR #379, opened 2016-10-27
Refuse to override builtin file type

bajtos:
Arguments of `{ type: 'file' }` are handled specially by strong-remoting and don't go through the usual TypeConverter path.

This causes problems in LoopBack applications with a model called "File", which registers a type converter for "file" (the name is case insensitive).

In this patch, we are modifying TypeRegistry to throw an error when the caller attempts to override the built-in "file" type.

Connect to strongloop/loopback#2554

> On the second thought, I think we should add `file` to the list of reserved types and disallow developers to create models called `File`, starting with LoopBack 3.0. It is already not possible to create models like `Object` or `Number`, therefore I think it makes sense to extend this rule to `File` too.
> I believe this will require the smallest amount of code changes and have the lowest maintenance costs in the future.

See also https://github.com/strongloop/strong-remoting/pull/370

@gunjpan or @deepakrkris please review
cc @fabien @ritch

### DIFF
--- a/lib/type-registry.js
+++ b/lib/type-registry.js
@@ -32,6 +32,10 @@ TypeRegistry.prototype.registerType = function(typeName, converter) {
 
   typeName = typeName.toLowerCase();
 
+  if (typeName === 'file') {
+    throw new Error(g.f('Cannot override built-in "{{file}}" type.'));
+  }
+
   if (typeName in this._types) {
     if (this._options.warnWhenOverridingType !== false)
       g.warn('Warning: overriding remoting type %s', typeName);
--- a/test/type-registry.test.js
+++ b/test/type-registry.test.js
@@ -1,0 +1,19 @@
+var expect = require('chai').expect;
+var TypeRegistry = require('../lib/type-registry');
+
+describe('TypeRegistry', function() {
+  var registry;
+  beforeEach(function() {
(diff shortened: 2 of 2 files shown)


lib/type-registry.js
```diff
@@ -32,6 +32,10 @@ TypeRegistry.prototype.registerType = function(typeName, converter) {
 
   typeName = typeName.toLowerCase();
 
+  if (typeName === 'file') {
+    throw new Error('Cannot override built-in "file" type.');
```
gunjpan: @bajtos : Please change this to throw a globalized error message as:
```javascript
  if(typename === 'file'){
    var eMsg = g.f('Cannot {{override}} built-in "{{file}}" type.');
    throw new Error(eMsg);
  }
```

gunjpan: Also, I don't see it being caught anywhere. So, I am assuming that it will bubble up and eventually handled. Correct?

bajtos: > Also, I don't see it being caught anywhere. So, I am assuming that it will bubble up and eventually handled. Correct?

It will bubble up and crash the app at startup, which is my intent. Let the user know about the problem immediately at startup, rather than later when some of their requests start failing.

bajtos:
@gunjpan fixed, PTAL again

PR #390, opened 2016-12-06
Rest-adapter errorHandler set content-type `json`

davidcheung:
the defaultHandler always res.sends an object
therefore make sense to always set response header as such

related to https://github.com/strongloop/loopback-component-storage/pull/179

### DIFF
--- a/test/rest.test.js
+++ b/test/rest.test.js
@@ -210,6 +210,26 @@ describe('strong-remoting-rest', function() {
         .end(done);
     });
 
+    // this is to verify scenarios such as file download
+    // where it sets content-type to image/jpeg, but file is 404 and
+    // response error body in strong-remoting(2.x) default error handler is always object
+    // see: https://github.com/strongloop/loopback-component-storage/pull/179
+    it('should use `content-type: application/json` for default errorHandler',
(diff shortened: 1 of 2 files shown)


test/rest.test.js
```diff
@@ -210,6 +210,24 @@ describe('strong-remoting-rest', function() {
         .end(done);
     });
 
+    it('should use `content-type: application/json` for default errorHandler',
```
bajtos: Could you please add a comment before this test case with a link to the github issue this test is reproducing?

test/rest.test.js
```diff
@@ -210,6 +210,24 @@ describe('strong-remoting-rest', function() {
...
+        cb(err);
+      });
+      objects.before(method.name, function(ctx, next) {
+        ctx.res.set('content-type', 'image/png');
+        next();
+      });
+
+      request(app).get(method.url)
+        .expect(500)
+        .expect('content-type', 'application/json; charset=utf-8', function(err) {
```
bajtos: You can simplify this line:

```js
.expect('content-type', '...', done);
```

davidcheung:
@slnode test please

PR #394, opened 2017-01-06
Upgrade eslint-config, improve npm scripts

bajtos:
- Add "npm run lint" script	
 - Use dot reporter in "npm test". We have more than 2000 tests now, listing names of all these tests is rather unpractical.
 - Upgrade eslint-config to 7.x, upgrade grunt-eslint to 19.x and add dirty-chai dev-dependency
to get rid of property-based assertions.

@superkhau PTAL


test/rest.test.js
```diff
@@ -1398,23 +1399,22 @@ describe('strong-remoting-rest', function() {
...
           }
         );
         request(app).get(method.classUrl)
           .set('Accept', 'application/xml')
           .set('Content-Type', 'application/json')
           .send()
-          .expect('Content-Type', /xml/)
+          .expect('Content-Type', /xml.*charset=utf-8/)
           .expect(200, function(err, res) {
-            expect(res).to.be.utf8;
```
bajtos: I had a WTF moment here. I think the intention was to assert that the response is in utf-8 encoding, which I am doing few lines above when checking the value of `Content-Type` header.

bajtos:
The best way for reviewing the diff is to ignore white-space changes - https://github.com/strongloop/strong-remoting/pull/394/files?w=1

bajtos:
@slnode test please

bajtos:
> Why default mocha to -R dot

With the default "list" reporter, `npm test` prints over 2000 lines to stdout, which I find annoying. `loopback` uses the "dot" reporter too, see https://github.com/strongloop/loopback/blob/4e6f9b978a50af82e2b8b461caf4d418ee439f43/Gruntfile.js#L92.

>  (ie. should we do it across the board or some specific reason this repo is different?).

To be honest, I don't care. I think there is no harm in using the "dot" reporter in all our projects, but at the same time I don't think it's worth the effort.

superkhau:
>...prints over 2000 lines to stdout, which I find annoying. 

Agreed lol, 2K lines seems like overkill.

>...no harm in using the "dot" reporter...

I agree with it's not worth the effort. Based on your point above though, I will do it for some repos that are absurdly long (ie. like juggler too) on as needed basis too. :bowing_man:

PR #400, opened 2017-03-14
Cache getRestMethodByName to improve performance of remote invocations

Traksewt:
This function in RestAdapter is slow due
to creating RestMethods each time.
This can be cached initially, and rechecked
if the entry was not found.

### Description
getRestMethodByName was identified as slow performing when syncing many items across many models. The reason is that each call will iterate through all classes, and create new RestMethods which is slow. 

This optimisation will create a name->RestMethod cache map which is initialised the first time it is called. This way, the look ups are fast. If there is a cache miss, then it will try to refresh the cache (if it hasn't been refreshed already).

#### Related issues
- None

### Checklist

### DIFF
--- a/lib/rest-adapter.js
+++ b/lib/rest-adapter.js
@@ -143,17 +143,43 @@ RestAdapter.prototype.invoke = function(method, ctorArgs, args, callback) {
   });
 };
 
-RestAdapter.prototype.getRestMethodByName = function(name) {
+/**
+ * creates the rest method by name cache map.
+ * @returns {Object} map of rest method name to rest method
+ */
+RestAdapter.prototype._createRestMethodByNameCache = function() {
+  var restMethodByNameMap = {};
   var classes = this.getClasses();
   for (var i = 0; i < classes.length; i++) {
     var restClass = classes[i];
     for (var j = 0; j < restClass.methods.length; j++) {
       var restMethod = restClass.methods[j];
-      if (restMethod.fullName === name) {
-        return restMethod;
-      }
+      restMethodByNameMap[restMethod.fullName] = restMethod;
     }
   }
+  return restMethodByNameMap;
+};
+
+RestAdapter.prototype.getRestMethodByName = function(name) {
+  var ret;
+  if (this._cachedRestMethodsByName) {
+    ret = this._cachedRestMethodsByName[name];
+  }
+
+  if (!ret) {
+    // Either the method was not found or the cache was not built yet
+    // If the method was not found, then let's rebuild the cache
+    // to see if there were any new methods added
+    this._cachedRestMethodsByName = this._createRestMethodByNameCache();
+    ret = this._cachedRestMethodsByName[name];
+  }
+
+  if (ret && !ret.sharedMethod.sharedClass.isMethodEnabled(ret.sharedMethod)) {
+    // The method was disabled after our cache was built
+    ret = undefined;
+  }
+
+  return ret;
 };
 
 /*!
--- a/test/rest-adapter.test.js
+++ b/test/rest-adapter.test.js
@@ -476,6 +476,61 @@ describe('RestAdapter', function() {
       return new RestAdapter(remotes);
     }
   });
+
+  describe('getRestMethodByName()', function() {
+    var SHARED_CLASS_NAME = 'testClass';
+    var METHOD_NAME = 'testMethod';
+    var FULL_NAME = SHARED_CLASS_NAME + '.' + METHOD_NAME;
+
+    var sharedClass, restAdapter;
+
+    beforeEach(givenSharedClassWithStaticMethod);
+    beforeEach(givenConnectedRestAdapter);
+
+    it('should find method by name', function() {
+      var restMethod = restAdapter.getRestMethodByName(FULL_NAME);
+      expect(restMethod).to.have.property('fullName', FULL_NAME);
+    });
+
+    it('should exclude methods disabled after the cache was built', function() {
+      // Get the rest method to trigger cache rebuild
+      restAdapter.getRestMethodByName(FULL_NAME);
+
+      sharedClass.disableMethodByName(METHOD_NAME);
+
+      var restMethod = restAdapter.getRestMethodByName(FULL_NAME);
+      expect(restAdapter.getRestMethodByName(FULL_NAME)).to.equal(undefined);
+    });
+
+    it('should find methods added after the cache was built', function() {
+      // Get the rest method to trigger cache rebuild
+      restAdapter.getRestMethodByName(FULL_NAME);
+
+      givenStaticSharedMethod('anotherMethod');
+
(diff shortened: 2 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

bajtos:
@slnode ok to test

bajtos:
Hi @Traksewt, thank you for the pull request. If my understanding is correct, then you are using strong-remoting to remotely invoke LoopBack server and this patch should make such invocations faster. Is that correct?

You proposed patch looks reasonably in general. My main concern is cache invalidation: in strong-remoting, it's possible to disable remote methods on the fly, it also may be possible to change remoting metadata too (at least in theory). I think the caching solution should be prepared to handle such situations, e.g. by invalidating the cache entry when `SharedClass.prototype.disableMethodByName` was called.

Thoughts?

lib/rest-adapter.js
```diff
@@ -143,17 +143,36 @@ RestAdapter.prototype.invoke = function(method, ctorArgs, args, callback) {
...
-    for (var j = 0; j < restClass.methods.length; j++) {
-      var restMethod = restClass.methods[j];
-      if (restMethod.fullName === name) {
-        return restMethod;
+/**
+ * gets or creates the rest method by name cache map.
+ * @param reload forces the map to refresh if true
+ * @returns {boolean} returns true if the cache was updated
+ */
+RestAdapter.prototype.getOrCreateRestMethodByNameMap = function(reload) {
```
bajtos: 1. Let's make this method private please - add `_` prefix to the method name.
2. Boolean arguments are difficult to understand - the line `this.getOrCreateRestMethodByNameMap(true);` below does not make the intent (reload the cache) clear.
3. I think this method is doing too much and should be simplified to rebuilding the cache only. Let the caller handle reload.

lib/rest-adapter.js
```diff
@@ -143,17 +143,36 @@ RestAdapter.prototype.invoke = function(method, ctorArgs, args, callback) {
...
+  var updated = false;
+  if (!this.restMethodByNameMap || reload) {
+    this.restMethodByNameMap = {};
+    var classes = this.getClasses();
+    for (var i = 0; i < classes.length; i++) {
+      var restClass = classes[i];
+      for (var j = 0; j < restClass.methods.length; j++) {
+        var restMethod = restClass.methods[j];
+        this.restMethodByNameMap[restMethod.fullName] = restMethod;
```
bajtos: The property should be private too. IMO, the name should include "cache" to make the intent more clear, e.g. `this._cachedRestMethods` or `this._cachedRestMethodsByName`.

test/rest-adapter.test.js
```diff
@@ -460,6 +460,23 @@ describe('RestAdapter', function() {
 
     });
 
+    it('should find RestMethod by name', function(done) {
+      var name = 'testClass.testMethod';
+      var _res;
```
bajtos: Is there any particular reason for choosing `_res` instead of `res`?

bajtos:
Last but not least, please rebase your feature branch on top of the latest `2.x`. I enabled coveralls code coverage in #402 to allow us to better review which parts of the new code may be missing test coverage.

test/rest-adapter.test.js
Traksewt: I just copy pasted from above. I will change all to just res.

lib/rest-adapter.js
Traksewt: makes sense.

lib/rest-adapter.js
Traksewt: yes this is better and cleaner.

test/rest-adapter.test.js
Traksewt: I see res is already defined at the top invoke() parent test. That is why all the children are using _res.

Traksewt:
Hi @bajtos. Our use case is that we are using the offline sync on a cordova based mobile app and syncing to a cloud server. As we are moving towards production, we have noticed some performance and scale issues with sync that we have been working through. 

I did the changes requested and added some more unit tests.

Traksewt:
Hi @bajtos I've made the changes. Please review.

lib/rest-adapter.js
```diff
@@ -143,17 +143,45 @@ RestAdapter.prototype.invoke = function(method, ctorArgs, args, callback) {
...
+    this._cachedRestMethodsByName = this._createRestMethodByNameCache();
+    cacheCreatedNow = true;
+  }
+  var ret = this._cachedRestMethodsByName[name];
+  if (!ret && !cacheCreatedNow) {
+    // check if it has been added recently
+    this._cachedRestMethodsByName = this._createRestMethodByNameCache();
+    ret = this._cachedRestMethodsByName[name];
+  }
+  if (ret && !ret.sharedMethod.sharedClass.isMethodEnabled(ret.sharedMethod)) {
```
bajtos: This changes the behaviour of `getRestMethodByName` beyond adding a transparent caching. What is the reasoning for that change? I'd like to leave it out of scope of this pull request.

lib/rest-adapter.js
```diff
@@ -143,17 +143,45 @@ RestAdapter.prototype.invoke = function(method, ctorArgs, args, callback) {
...
+    ret = null;
+  }
+  return ret;
+};
+
+RestAdapter.prototype.invalidateRestMethodCache = function() {
```
bajtos: This method is not used anywhere AFAICT. Is the intention to require application code to call this method? That would be difficult to use IMO, because LoopBack users will not know they have to call this method manually.

lib/rest-adapter.js
bajtos: Ah, is this needed to support `should not find RestMethod if it is disabled` test?

lib/rest-adapter.js
bajtos: Since you are already checking for disabled methods on L175 above (where I commented because I was confused), I think there is no need to have `invalidateRestMethodCache` and I am proposing to remove this method.

Traksewt: OK removed.

lib/rest-adapter.js
Traksewt: Right, I was adding disabled support from SharedClass as discussed above.

Traksewt:
@bajtos changes done. please review.

bajtos:
@Traksewt thanks. To speed up the review process, I have simplified and clean up your code in 634f646. PTAL and let me know if you are ok with my changes.

Traksewt:
@bajtos sure. The tests look neater.

bajtos:
Great! I have squashed all commits into a single one and will land this pull request after CI builds pass.

bajtos:
Landed, thank you for the contribution!

PR #408, opened 2017-05-26
Support custom transport/adapter implementations

pierissimo:
### Description
Implemented a mechanism allowing strong-remoting users to provide their own transport/adapter implementations.

#### Related issues

- #141 

### Checklist

- [X] New tests added or existing tests modified to cover all changes
- [X] Code conforms with the [style

### DIFF
--- a/lib/remote-objects.js
+++ b/lib/remote-objects.js
@@ -89,13 +89,13 @@ RemoteObjects.extend = function(exports) {
 /**
  * Create a handler from the given adapter.
  *
- * @param {String} name Adapter name
+ * @param {String} nameOrClass Adapter name or provided adapter class
  * @param {Object} options Adapter options
  * @return {Function}
  */
 
-RemoteObjects.prototype.handler = function(name, options) {
-  var Adapter = this.adapter(name);
+RemoteObjects.prototype.handler = function(nameOrClass, options) {
+  var Adapter = this.adapter(nameOrClass);
   var adapter = new Adapter(this, options);
   var handler = adapter.createHandler();
 
@@ -160,12 +160,40 @@ RemoteObjects.prototype.invoke = function(method, ctorArgs, args, callback) {
 
 /**
  * Get an adapter by name.
- * @param {String} name The adapter name
+ * @param {String} nameOrClass The adapter name or provided class
  * @return {Adapter}
  */
 
-RemoteObjects.prototype.adapter = function(name) {
-  return require('./' + name + '-adapter');
+RemoteObjects.prototype.adapter = function(nameOrClass) {
+  if (typeof nameOrClass === 'function') {
+    // Throws an error if the provided adapter doesn't meets the requirements
+    this.validateAdapter(nameOrClass);
+
+    return nameOrClass;
+  } else {
+    return require('./' + nameOrClass + '-adapter');
+  }
+};
+
+RemoteObjects.prototype.getRequiredAdapterMethods = () => ['createHandler'];
+
+RemoteObjects.prototype.validateAdapter = function(Adapter) {
+  const requiredMethods = this.getRequiredAdapterMethods();
+
+  const notImplementedMethods = requiredMethods.filter(
+      methodString => typeof Adapter.prototype[methodString] !== 'function'
+  );
+
+  if (notImplementedMethods.length > 0) {
+    throw new Error(g.f(`
+        Invalid adapter class. 
+        The provided adapter class does not implement 
+        the following required methods: %s`,
+        notImplementedMethods.join(', ')
+    ));
+  }
+
+  return Adapter;
 };
 
 /**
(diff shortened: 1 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

lib/remote-objects.js
```diff
@@ -161,11 +161,34 @@ RemoteObjects.prototype.invoke = function(method, ctorArgs, args, callback) {
...
 
-RemoteObjects.prototype.adapter = function(name) {
-  return require('./' + name + '-adapter');
+RemoteObjects.prototype.adapter = function(nameOrClass) {
+  if (typeof nameOrClass === 'function') {
+    var requiredMethods = this.getRequiredAdapterMethods();
+    var meetRequirements = this.checkAdapterRequirements(nameOrClass, requiredMethods);
+    if (!meetRequirements) {
+      throw new Error(g.f('Invalid adapter class'));
+    }
```
bajtos: I am proposing to move the validation into a new method, e.g. `this.validateAdapter(adapter)`.

Also in my experience, a generic error like "Invalid adapter class" makes it difficult for users to figure out what's the problem and how to fix it. A better solution is to change the code checking adapter requirements and return either a full error, or at least an error-message string, instead of current true/false value.

test/remote-objects.test.js
```diff
@@ -0,0 +1,37 @@
...
+var factory = require('./helpers/shared-objects-factory.js');
+var RemoteObjects = require('../');
+var RestAdapter = require('../lib/rest-adapter');
+function NOOP() {};
+
+describe('RemoteObjects', function() {
+  var remotes;
+  beforeEach(function() { remotes = RemoteObjects.create(); });
+
+  describe('RemoteObjects.methods()', function() {
```
bajtos: The test is invoking `RemoteObjects.handler`, not `method`.

pierissimo:
thanks for the review, I just pushed the code with the changes you mentioned.

lib/remote-objects.js
```diff
@@ -160,12 +160,46 @@ RemoteObjects.prototype.invoke = function(method, ctorArgs, args, callback) {
...
+  var requiredMethods = this.getRequiredAdapterMethods();
+
+  var notImplementedMethods = requiredMethods.reduce(
+      function(notImplementedMethods, methodString) {
+        if (typeof Adapter.prototype[methodString] !== 'function') {
+          return notImplementedMethods.concat(methodString);
+        }
+
+        return notImplementedMethods;
+      }, []);
```
bajtos: I think this can be simplified by using `filter`:

```js
const notImplementedMethods = requireMethods.filter(
  methodString => typeof Adapter.prototype[methodString] !== 'function')
```

Note that since 3.0 release, we target Node 4+, thus we can (and should be) using ES6 constructs like `const` and arrow functions.

See also:
 - http://loopback.io/doc/en/contrib/style-guide.html#variable-declarations
 - http://loopback.io/doc/en/contrib/style-guide.html#arrow-functions

lib/remote-objects.js
```diff
@@ -160,12 +160,46 @@ RemoteObjects.prototype.invoke = function(method, ctorArgs, args, callback) {
...
+
+  if (notImplementedMethods.length > 0) {
+    throw new Error(g.f(
+        'Invalid adapter class. ' +
+        'The provided adapter class doesn\'t implement ' +
+        'the required methods: ' + notImplementedMethods.join(',')
```
bajtos: `g.f` behaves like `util.format`, with the difference that the first argument is translated before performing string replacements on `%s` and similar. The variable part of the message should be passed as the second argument.

AFAIK, there is a typographic convention to add a space after a comma, therefore the `join()` separator should be `', '`.

I am reading "the required methods"  as "all required methods", whereas the error shows only the methods that are missing. I think "the following required methods" may be a better formulation.

```js
throw new Error(g.f(
  'Invalid adapter class. ' +
    'The provided adapter class does not implement ' +
    'the following required methods: %s',
  notImplementedMethods.join(', ')));
```

lib/remote-objects.js
pierissimo: Great. I was not aware of

lib/remote-objects.js
pierissimo: I see. Fixing

pierissimo:
Fixed the problems you mentioned.

Arrow functions are not applicable to the .prototype syntax, because they cause an incorrect context binding.
The 'class' syntax would be preferable.

bajtos:
@slnode ok to test

bajtos:
Landed, thank you for the contribution!

When you have your Primus adapter ready, could you please submit a pull request to add it to Community Projects on loopback.io, so that we can help you to publicize it? See http://loopback.io/doc/en/community/How-to-add-a-project.html

pierissimo:
@bajtos 
thank you, I will do!

PR #410, opened 2017-07-07
Add support for app-wide "normalizeHttpPath"

lehni:
### Description

Setting `"normalizeHttpPath"` in `confing.json` currently has no effect on models:

```json
"remoting": {
  "rest": {
    "normalizeHttpPath": true
  }
}
```

This PR addresses this issue by implementing a proposal that @ritch has mentioned in a comment here:

https://github.com/strongloop/strong-remoting/blob/master/lib/shared-class.js#L64

#### Related issues

strongloop/loopback#608

### DIFF
--- a/lib/rest-adapter.js
+++ b/lib/rest-adapter.js
@@ -30,6 +30,7 @@ var async = require('async');
 var HttpInvocation = require('./http-invocation');
 var HttpContext = require('./http-context');
 var strongErrorHandler = require('strong-error-handler');
+var inflection = require('inflection');
 
 var json = bodyParser.json;
 var urlencoded = bodyParser.urlencoded;
@@ -69,20 +70,31 @@ RestAdapter.createRestAdapter = function(remotes) {
  * Get the path for the given method.
  */
 
-RestAdapter.prototype.getRoutes = getRoutes;
-function getRoutes(obj) {
-  var routes = obj.http;
+RestAdapter.prototype.getRoutes = function(obj) {
+  return getRoutes(obj, this.options);
+};
 
+function getRoutes(obj, options) {
+  var routes = obj.http;
   if (routes && !Array.isArray(routes)) {
     routes = [routes];
   }
 
-  // overidden
+  // Options of obj (e.g. sharedClass) take precedence over options of adapter
+  const sharedClass = obj.sharedClass || obj;
+  const classOptions = sharedClass.options;
+  var normalize = classOptions && classOptions.normalizeHttpPath;
+  if (normalize === undefined && options) {
+    normalize = options.normalizeHttpPath;
+  }
+  var toPath = normalize ? normalizeHttpPath : untransformedPath;
+
+  // overridden
   if (routes) {
     // patch missing verbs / routes
     routes.forEach(function(r) {
       r.verb = String(r.verb || 'all').toLowerCase();
-      r.path = r.path || ('/' + obj.name);
+      r.path = toPath(r.path || ('/' + obj.name));
     });
   } else {
     if (obj.name === 'sharedCtor') {
@@ -94,12 +106,27 @@ function getRoutes(obj) {
       // build default route
       routes = [{
         verb: 'all',
-        path: obj.name ? ('/' + obj.name) : '',
+        path: obj.name ? toPath('/' + obj.name) : '',
       }];
     }
   }
 
   return routes;
+}
+
+/**
+ * Normalize HTTP path.
+ */
+function normalizeHttpPath(path) {
+  if (typeof path !== 'string') return;
+  return path.replace(/[^\/]+/g, function(match) {
+    if (match.indexOf(':') > -1) return match; // skip placeholders
+    return inflection.transform(match, ['underscore', 'dasherize']);
+  });
+}
+
+function untransformedPath(path) {
+  return path;
 }
 
 RestAdapter.prototype.connect = function(url) {
@@ -122,6 +149,10 @@ RestAdapter.prototype.invoke = function(method, ctorArgs, args, callback) {
 
   var remotes = this.remotes;
   var restMethod = this.getRestMethodByName(method);
+  if (!restMethod) {
+    return callback(new Error(g.f('Cannot invoke unkown method: %s', method)));
+  }
+
   var invocation = new HttpInvocation(
     restMethod, ctorArgs, args, this.connection, remotes.auth, this.typeRegistry
   );
@@ -546,16 +577,17 @@ RestAdapter.prototype.allRoutes = function() {
 };
 
 RestAdapter.prototype.getClasses = function() {
-  return this.remotes.classes(this.options).map(function(c) {
-    return new RestClass(c);
+  return this.remotes.classes(this.options).map(c => {
+    return new RestClass(c, this.options);
   });
 };
 
-function RestClass(sharedClass) {
+function RestClass(sharedClass, adapterOptions) {
   nonEnumerableConstPropery(this, 'sharedClass', sharedClass);
 
   this.name = sharedClass.name;
-  this.routes = getRoutes(sharedClass);
+  this.options = adapterOptions;
+  this.routes = getRoutes(sharedClass, this.options);
 
   this.ctor = sharedClass.sharedCtor &&
     new RestMethod(this, sharedClass.sharedCtor);
@@ -586,7 +618,7 @@ function RestMethod(restClass, sharedMethod) {
   this.notes = sharedMethod.notes;
   this.documented = sharedMethod.documented;
 
-  var methodRoutes = getRoutes(sharedMethod);
+  var methodRoutes = getRoutes(sharedMethod, restClass.options);
   if (sharedMethod.isStatic || !restClass.ctor) {
     this.routes = methodRoutes;
   } else {
--- a/lib/shared-class.js
+++ b/lib/shared-class.js
@@ -20,7 +20,7 @@ module.exports = SharedClass;
 var debug = require('debug')('strong-remoting:shared-class');
 var util = require('util');
 var inherits = util.inherits;
-var inflection = require('inflection');
+var extend = util._extend;
 var SharedMethod = require('./shared-method');
 var assert = require('assert');
 
@@ -36,14 +36,13 @@ var assert = require('assert');
  */
 
 function SharedClass(name, ctor, options) {
-  options = options || {};
   this.name = name || ctor.remoteNamespace;
   this.ctor = ctor;
+  this.options = options;
   this._methods = [];
   this._resolvers = [];
   this._disabledMethods = {};
   var http = ctor && ctor.http;
-  var normalize = options.normalizeHttpPath;
 
   var defaultHttp = {};
   defaultHttp.path = '/' + this.name;
@@ -54,17 +53,9 @@ function SharedClass(name, ctor, options) {
     if (http.length === 0) {
       http.push(defaultHttp);
     }
-    if (normalize) {
-      this.http.forEach(function(h) {
-        h.path = SharedClass.normalizeHttpPath(h.path);
-      });
-    }
   } else {
     // set http.path using the name unless it is defined
-    // TODO(ritch) move http normalization from adapter.getRoutes() to a
-    // better place... eg SharedMethod#getRoutes() or RestClass
     this.http = util._extend(defaultHttp, http);
-    if (normalize) this.http.path = SharedClass.normalizeHttpPath(this.http.path);
   }
 
   if (typeof ctor === 'function' && ctor.sharedCtor) {
@@ -76,18 +67,6 @@ function SharedClass(name, ctor, options) {
   }
   assert(this.name, 'must include a remoteNamespace when creating a SharedClass');
 }
-
-/**
- * Normalize HTTP path.
- */
-
-SharedClass.normalizeHttpPath = function(path) {
-  if (typeof path !== 'string') return;
-  return path.replace(/[^\/]+/g, function(match) {
-    if (match.indexOf(':') > -1) return match; // skip placeholders
-    return inflection.transform(match, ['underscore', 'dasherize']);
-  });
-};
 
 /**
  * Get all shared methods belonging to this shared class.
--- a/test/shared-class.test.js
+++ b/test/shared-class.test.js
@@ -11,6 +11,7 @@ var expect = require('chai').expect;
 var SharedClass = require('../lib/shared-class');
 var factory = require('./helpers/shared-objects-factory.js');
 var RemoteObjects = require('../');
+var RestAdapter = require('../lib/rest-adapter');
 function NOOP() {};
 
 describe('SharedClass', function() {
@@ -31,7 +32,11 @@ describe('SharedClass', function() {
 
     it('fills http.path using a normalized path', function() {
       var sc = new SharedClass('SomeClass', SomeClass, {normalizeHttpPath: true});
-      expect(sc.http.path).to.equal('/some-class');
+      var remotes = RemoteObjects.create();
+      remotes.addClass(sc);
+      var classes = new RestAdapter(remotes).getClasses();
+      expect(classes[0]).to.have.property('routes')
+        .eql([{path: '/some-class', verb: 'all'}]);
     });
 
     it('does not require a sharedConstructor', function() {
--- a/test/jsonrpc.test.js
+++ b/test/jsonrpc.test.js
@@ -77,7 +77,7 @@ describe('strong-remoting-jsonrpc', function() {
           });
         };
 
-        var productClass = new SharedClass('product', Product, {});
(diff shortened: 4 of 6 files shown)


slnode:

slnode:

slnode:

slnode:

lehni:
Oops, I will look into the failing test and fix ASAP.

bajtos:
@slnode ok to test

lib/rest-adapter.js
```diff
@@ -586,7 +609,7 @@ function RestMethod(restClass, sharedMethod) {
   this.notes = sharedMethod.notes;
   this.documented = sharedMethod.documented;
 
-  var methodRoutes = getRoutes(sharedMethod);
+  var methodRoutes = getRoutes(sharedMethod, options);
```
bajtos: I am proposing to read `options` from `restClass` argument (or `this.restClass`), thus keeping the number of arguments of `RestMethod` smaller.

test/shared-class.test.js
```diff
@@ -29,11 +29,6 @@ describe('SharedClass', function() {
       expect(sc.http.path).to.equal('/some');
     });
 
-    it('fills http.path using a normalized path', function() {
-      var sc = new SharedClass('SomeClass', SomeClass, {normalizeHttpPath: true});
```
bajtos: We need to preserve this behaviour for backwards compatibility (existing applications should keep working after upgrading to the new version).

With two places where to configure `normalizeHttpPath`, we need to decide which one takes priority - model-level or app-level configuration?

@raymondfeng @ritch @superkhau do you happen to remember any example of a config option that can be set both on app-level and model-level? I think we should make `normalizeHttpPath` consistent with the rest of the existing options.

superkhau: I do not recall any off the top of my head.

bajtos: In that case, let's prioritize model-level setting over app-level setting. In other words, if application sets `normalizeHttpPath: false`, but model sets `normalizeHttpPath: true`, then the path should be normalized.

lehni: As far as I can tell, so far, the app-level setting had no effect whatsoever, and only the model level was actually sort of working, but only one level deep: nested routes were reverting to the default of not normalizing the paths, ignoring the model-level setting (as well as the app-level oe).

lib/rest-adapter.js
lehni: That makes sense. But what about merging both?

lehni: Although, when I look at the two places where `new RestMethod()`is called, they both receive the exact same `options` value as the `RestClass()` constructor from which they are called. The same value is passed on to `getRoutes()` also to determine the value of `this.routes` there. I don't see the `options` being stored on the `RestClass` instance: 

https://github.com/lehni/strong-remoting/blob/master/lib/rest-adapter.js#L577-L591

test/shared-class.test.js
lehni: @bajtos regarding preserving backwards compatibility: Is any app constructing `SharedClass` instances directly? The functionality is preserved, just on the level of the `RestAdapter` where it really belongs, and there are unit tests for that... I don't think it makes sense to transform the http routes on the `SharedClass` level, and that's what @ritch's original comment was about, I believe.

lib/rest-adapter.js
bajtos: Usually, when a constructor accepts an `options` argument, then we store it in the class instances, so that `options` can be accessed from any member method. See [RemoteObjects constructor](https://github.com/strongloop/strong-remoting/blob/5b89fb2ab79a77c5b4d25b4009e8033951eb4852/lib/remote-objects.js#L63). OTOH, apparently SharedClass and SharedMethod don't store options directly, so we are not consistent in this.

> I don't see the `options` being stored on the `RestClass` instance.

`options` is a new argument added by your pull request. Since you are adding it, you can also store it in the instance.

This is not a big deal though. If you prefer to keep passing options via function arguments instead of `this.options`, then I can live with that too.

test/shared-class.test.js
bajtos: > As far as I can tell, so far, the app-level setting had no effect whatsoever, and only the model level was actually sort of working, but only one level deep: nested routes were reverting to the default of not normalizing the paths, ignoring the model-level setting (as well as the app-level one).

Yes, your assessment looks correct to me.

>  Is any app constructing `SharedClass` instances directly? 

Whenever you create a new model in your LoopBack application, a new `SharedClass` instance is created for this model. See [lib/model.js#L206-L211](https://github.com/strongloop/loopback/blob/ca3a21ddd5f4e29d4996858b126ca8b53858ff1c/lib/model.js#L206-L211) in strongloop/loopback.

Here is the use case we need to preserve:
 - create a LoopBack app
 - add a `FooBar` model
 - configure the model to `normalizeHttpPath` via `common/models/foobar.json`
 - observe that the model is mounted at `/api/foo-bar`

Such application should keep working (with `/api/foo-bar`) even after upgrading to your new version of strong-remoting, without any configuration changes needed. That's our understanding of backwards compatibility.

> I don't think it makes sense to transform the http routes on the SharedClass level, and that's what @ritch's original comment was about, I believe.

I am afraid I don't know what do you mean by @ritch's original comment, could you please paste a link? You can get the link by clicking on a date in the comment header.

To me, it makes a lot of sense to allow LoopBack users to configure normalized HTTP paths at multiple levels:

 - app-wide (as you are implementing here)
 - individually for each model (as is implemented now, with nested routes not normalized)

lib/rest-adapter.js
lehni: Haha, you're right about me adding it. Apologies, I've been commenting from a mini-holiday, brain half-absent. I am happy to implement it in whatever way you deem best. I will update the PR accordingly. Is force pushing a new, single commit the best way to do so?

test/shared-class.test.js
lehni: This is the comment I meant: 

https://github.com/strongloop/strong-remoting/blob/master/lib/shared-class.js#L64

And I agree that we need to preserve this. We should also have unit tests for it. Is it easy to achieve, to test for a full app scenario?

lib/rest-adapter.js
bajtos: > Is force pushing a new, single commit the best way to do so?

Either that, or simply keep adding more commits to this feature branch.

test/shared-class.test.js
bajtos: > Is it easy to achieve, to test for a full app scenario?

Not really. Here in strong-remoting, we usually write a test that's exercising the same call/code path as LoopBack (a full app) uses. In this particular case, it should be enough to keep this unit test that you originally proposed to remove. (I.e. revert this removal.)

lib/rest-adapter.js
lehni: I've addressed this now in the new version that I pushed.

test/shared-class.test.js
lehni: I believe I found the right solution now. Please review the updated PR. The mentioned unit test is still there, but it needs to test the normalization through a `RestAdapter`, as the logic to normalize the HTTP paths is there now. I've adjusted the test accordingly. From my tests, all these scenarios of different values for `normalizeHttpPath` work as expected now:

| `config.json`  | `model.json` | Result |
| -------------- | ------------- | ------ |
| `undefined`  | `undefined`  | `ModelName/relationshipName` |
| `undefined`  | `false` | `ModelName/relationshipName` |
| `undefined`  | `true`  | `model-name/relationshipName` |
| `false`  | `undefined`  | `ModelName/relationshipName` |
| `true`  | `undefined`  | `model-name/relationship-name` |
| `false`  | `false`  | `ModelName/relationshipName` |
| `false`  | `true`  | `model-name/relationshipName` |
| `true`  | `false`  | `ModelName/relationship-name` |
| `true`  | `true`  | `model-name/relationship-name` |

test/shared-class.test.js
```diff
@@ -31,7 +32,11 @@ describe('SharedClass', function() {
 
     it('fills http.path using a normalized path', function() {
       var sc = new SharedClass('SomeClass', SomeClass, {normalizeHttpPath: true});
-      expect(sc.http.path).to.equal('/some-class');
+      var remotes = RemoteObjects.create();
+      remotes.addClass(sc);
+      var classes =  new RestAdapter(remotes).getClasses();
```
bajtos: two spaces after `=`, please remove one of them.

bajtos:
As I was thinking about this pull request, I realized we need to review the affect on the REST client that is part of strong-remoting. In essence, one can use the same remoting metadata used by the server to build a client which is capable of invoking the defined remote methods. See the tests in [rest.test.js](https://github.com/strongloop/strong-remoting/blob/5b89fb2ab79a77c5b4d25b4009e8033951eb4852/rest.test.js#L2348), starting at line 2348.

I think that by moving the code normalizing HTTP paths into RestAdapter, we will break the client side, because it will see the original (not normalized) paths.

Could you please check?

It would be great to add few tests to verify how `normalizeHttpPath` is supported by the client.

lehni:
I have never used with the client, and I cannot see the file you're linking to (I get an error). Could you help me with these tests? This is beyond what I understand and use of LoopBack, currently...

lehni:
@bajtos I've found the file and looked into this now : )

The problem I am seeing is that the tests you point to use `givenSharedStaticMethod()`, which doesn't even create a `SharedClass` instance, it just creates a object literal `remotes.testClass = {testMethod: fn}`. i don't think there would be any normalization either way then?

I also didn't see any tests that were actually using the paths to invoke the methods. They all use `objects.invoke()` with `method.name` directly... I do't really know how I would have to write these tests : /

bajtos:
> The problem I am seeing is that the tests you point to use givenSharedStaticMethod(), which doesn't even create a SharedClass instance. It just creates an object literal:
`remotes.testClass = {testMethod: fn}`

This is the old (legacy) way of defining remote methods, we are supporting it for backwards compatibility.

> I don't think there would be any normalization either way then?

Well, there will be normalization when it's configured at the adapter level.

> I also didn't see any tests that were actually using the paths to invoke the methods. They all use `objects.invoke()` with method.name directly... I do't really know how I would have to write these tests : /

Yeah, I see what you mean. I'll try to find some time next week to investigate this and provide you with an example of such test.

lehni:
I think even currently, if you configure normalization on the server in a model, the client side would struggle with it... I've looked into the code, which is fairly hard to understand, and couldn't grasp how I could add support to the client side. I will try more again later...

lehni:
I'm just realising that my latest change broke app-wide normalization. While the explorer lists the correct routes, the unnormalized appear to get installed and used. I am looking into what's causing this and will fix it. We also need tests for app-wide normalization for sure...

lehni:
I have managed to track this down now to `RemoteObjects.prototype.classes(options)`, where `this._classes` contained previously created `SharedClass` instances which didn't receive the passed `options`.

I have now added `SharedClass.prototype.setOptions(options)` which merges these into the already set options, and this has addressed the problem. 

I've pushed the new changes.

lib/shared-class.js
```diff
@@ -78,15 +69,17 @@ function SharedClass(name, ctor, options) {
...
-SharedClass.normalizeHttpPath = function(path) {
-  if (typeof path !== 'string') return;
-  return path.replace(/[^\/]+/g, function(match) {
-    if (match.indexOf(':') > -1) return match; // skip placeholders
-    return inflection.transform(match, ['underscore', 'dasherize']);
-  });
+SharedClass.prototype.setOptions = function(options) {
+  if (options) {
+    this.options = !this.options ? options :
+      extend(extend({}, this.options), options);
```
bajtos: Uff, this is rather complex solution. How about using Object.assign?


```js
SharedClass.prototype.setOptions = function(options) {
  this.options = Object.assign({}, this.options, options);
}
```

bajtos:
@lehni I added two tests for invoking shared methods from the client side in 56a17b2. While working on those two tests, I discovered some problems in your proposed implementation, I have fixed them along the way. I feel this pull request really needs more unit test to give us more confidence about the correctness of the implementation. (See https://github.com/strongloop/strong-remoting/pull/410#pullrequestreview-50256361 for two examples.)

lib/rest-adapter.js
```diff
@@ -69,20 +70,31 @@ RestAdapter.createRestAdapter = function(remotes) {
...
 
+function getRoutes(obj, options) {
+  var routes = obj.http;
   if (routes && !Array.isArray(routes)) {
     routes = [routes];
   }
 
-  // overidden
+  // Options of obj (e.g. sharedClass) take precedence over options of adapter
+  const sharedClass = obj.sharedClass || obj;
```
bajtos: `obj` can be either a `SharedClass` instance (I think that's for constructors?) or a `SharedMethod` instance. In the second case we need to read the options from `SharedClass`.

lib/rest-adapter.js
```diff
@@ -571,16 +577,16 @@ RestAdapter.prototype.allRoutes = function() {
 };
 
 RestAdapter.prototype.getClasses = function() {
-  return this.remotes.classes(this.options).map(function(c) {
-    return new RestClass(c);
+  return this.remotes.classes(this.options).map(c => {
+    return new RestClass(c, this.options);
```
bajtos: We must not store `RestAdapter` options inside `SharedClass` instance, because that breaks to logic merging class-level with app/adapter-level settings. Instead, REST-related options should be stored inside `RestClass`.

lehni: Is it OK to use ES6 style in the code? I thought it was still all written in ES5 manner...

lehni: Hmm but this now breaks the app-level setting of `normalizeHttpPath` again in my app. : / I really don't understand enough about all the involved classes, use cases and involved complexities to fix this properly. I will try to figure out what's missing now...

bajtos: Most of our packages allow ES6 style (as supported by Node.js 4.x) now. See [engines in package.json](https://github.com/strongloop/strong-remoting/blob/96042273024b8b8cc17d2576d6967cd4b089209a/package.json#L18) and http://loopback.io/doc/en/contrib/style-guide.html.

In general, we are keeping existing code as it is, but newly added lines can/should leverage ES6  syntax.

lib/shared-class.js
lehni: I wasn't sure if it was OK to use `Object.assign()` or not. But that code is now gone anyway!

lehni:
@bajtos your changes look good to me, thanks! I've tested them and they work for me, including nested routes. I am not sure I can be of much assistance with those additional unit tests unfortunately, as this is a bit beyond my understanding of all the related parts... What's the next step for this PR now?

bajtos:
It would be nice to have few more tests to check how app-level and class-level settings interact (which one has priority). Unfortunately I don't have bandwidth to add them myself. I don't want to block progress on this feature, and considering that there are at least some tests, I think it's ok to land this pull request.

I checked why the downstream "loopback" build is failing, the problem is not related to this patch.

Here is what I am going to do next: rebase the changes on top of the latest master, squash commits into a single one and then wait one more time for CI results before landing.

lehni:
I am happy to add more tests for this... Shall I do so in a separate PR? I could work on this right now, actually. So maybe you want to wait with the merge?

lehni:
Just realised again that those tests are in https://github.com/strongloop/loopback/issues/3527, not here, so I can add more there anyhow. I will work on it now.

bajtos:
> I can add more there anyhow. I will work on it now.

That's awesome, I am happy to give this pull request more time.

You will need to hard-reset your local clone because I did `git push -f` into your master branch. Here is one way how to do that:

```sh
# save changes that are not committed yet
$ git stash

$ git fetch origin
$ git reset --hard origin/master

# get back stashed changes
$ git stash pop
```

lehni:
@bajtos I added an additional test for HTTP normalization precedence now:
https://github.com/strongloop/loopback/pull/3527/commits/7fd5546e892a5253e0e88bf651117a180da5bf65

Unfortunately, this one is currently failing! The way things are implemented at the moment in this PR here, the app settings take precedence over the Model settings. I don't think this is right.

bajtos:
> I added an additional test for HTTP normalization precedence now:
strongloop/loopback@7fd5546
>
> Unfortunately, this one is currently failing! The way things are implemented at the moment in this PR here, the app settings take precedence over the Model settings. I don't think this is right.

That's weird. I agree with you that app settings should not be taking precedence over the Model settings.

lehni:
@bajtos ok, trying to figure this one out...

lehni:
@bajtos ok my bad! I had my local branch in a funny state and didn't catch it. All tests pass with the latest state of this PR. It's good to go!

bajtos:
Landed, thank you for the contribution! 🎉

bajtos:
Released in `strong-remoting@3.4.0`

PR #417, opened 2017-08-10
fix: destroy stream result on client close

perkd:
when SSE connection is closed, destroy PassThrough stream created in createChangeStream()

### DIFF
--- a/lib/http-context.js
+++ b/lib/http-context.js
@@ -417,6 +417,17 @@ HttpContext.prototype.respondWithEventStream = function(stream) {
   stream.on('end', function() {
     client.send({event: 'end', data: 'null'});
   });
+
+  if (stream.destroy) {
+    // ReadableStream#destroy() was added in Node.js 8.0.0
+    // In earlier versions, some kinds of streams are already
+    // providing this method, which allows us to correctly cancel them.
+    // For streams that do not provide this method, there
+    // is no other reliable way for closing them prematurely.
+    client.once('close', function() {
+      stream.destroy();
+    });
+  }
 };
 
 /**
--- a/test/streams.js
+++ b/test/streams.js
@@ -70,7 +70,7 @@ describe('strong-remoting', function() {
 describe('a function returning a ReadableStream', function() {
   var Readable = require('stream').Readable;
   var remotes = RemoteObjects.create();
-  var streamClass, server, app;
+  var streamClass, server, app, streamClosed;
 
   before(function(done) {
     var test = this;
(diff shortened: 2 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

bajtos:
@slnode ok to test

bajtos:
Hello @perkd, thank you for the pull request.

Before we can land your patch, we need you to sign our Contributor's License Agreement - see https://cla.strongloop.com/agreements/strongloop/strong-remoting

The changes you are proposing make sense to me, but I would like to understand the wider context. What happens now, when the change stream is not destroyed? Are we leaking memory? Something else?

Could you please add a test to verify this fix and prevent regressions in the future?

perkd:
Hi @bajtos,

CLA signed.
Currently:
1. a new PassThrough stream is created every time a client request a change stream
2. this PassThrough stream is not destroyed when the client closes (or timeout) the connection
3. event handlers are not released: Model.observe('after save', ...) & Model.observe('after delete', ...)

Over time, especially under high turnover of change streams, (2) & (3) will lead to memory leak?

bajtos:
@perkd thank you for the comment, your explanation makes sense.

So, is it possible to write a test that will close the SSE client (I suppose by closing the response stream?) and then verify that the original stream returned by the remote method was closed too?

I am afraid the CLA check is still red, see "Some checks were not successful" below. Could you please try to sign the CLA again?

perkd:
Strange:

screen shot 2017-08-18 at 10 15 25 pm

perkd:
@bajtos, I'll need some time to figure out a simple way for the SSE client part.

bajtos:
@perkd oh. so the commit says it has been created by @youngtt, so we need that person to sign the  CLA using their GitHub account.

bajtos:
> I'll need some time to figure out a simple way for the SSE client part.

No worries. I think the test should use `http.request()` to call a test endpoint returning a change stream, and then call `response.destroy()` without waiting for the data. We should not need to worry about SSE.

youngtt:
@bajtos, CLA signed. My bad :)

youngtt:
@bajtos here's the test script:

`
"use strict";
const http = require("http");
var app = require("../server/server");
app.start();

describe('PersistedModel.createChangeStream', function() {
	before(function() {
		var myModel = app.registry.createModel('myModel');
		app.model(myModel, {
			dataSource: app.registry.memory(),
			public: true
		});
		myModel.original = myModel.createChangeStream;
	});

	it('destroy stream on disconnect by client', function() {
		var myModel = app.models.myModel;
		myModel.createChangeStream = function(options, cb) {
			myModel.original(options, function handler(err, changes) {
				changes.on('close', function() {
					console.log('--- stream (closed) ---');
				});
				cb(null, changes);
			});
		};
	});
	it('connect to changeStream and disconnect', function(done) {
		var req = http.get({
			host: 'localhost',
			port: 3000,
			path: '/api/myModels/change-stream'
		}, function(res) {
			res.socket.end();
			done();
		});
	});
});
`

bajtos:
Hello @perkd, thank you for the update. Your test unfortunately did not work out of the box, but it gave me an idea how to write a test that would work - see 7a5b2f7. Please take a look at my proposed test and let me know if you are ok to get it landed under your name :)

youngtt:
@bajtos great!  thanks for the help :)

bajtos:
Apparently `ReadableStream#destroy` was added in Node.js 8.0.0 only - see https://github.com/nodejs/readable-stream/issues/124. In 5f32093, I added a check to destroy the stream only if `destroy` is available, and reworked the new test accordingly.

bajtos:
Landed, thank you for your contribution! 🎉

PR #421, opened 2017-09-14
Use lb access token

Traksewt:
### Description
Allow the rest connector to call other loopback services by passing on the user authorization from the options where available. This is the fallback if no other authorization is specified.

This allows loopback servers to communicate securely with other loopback servers by passing on the auth token. Useful for using loopback as microservices.

#### Related issues
- connect to strongloop/loopback-connector-rest#89
- connect to strongloop/loopback-connector-remote#3

### Checklist

### DIFF
--- a/lib/remote-objects.js
+++ b/lib/remote-objects.js
@@ -93,9 +93,7 @@ RemoteObjects.extend = function(exports) {
 
 RemoteObjects.prototype.handler = function(name, options) {
   var Adapter = this.adapter(name);
-  // send the set options for this remote object and
-  // the dynamic options passed in to create the adapter
-  var adapter = new Adapter(this, Object.assign({}, this.options, options));
+  var adapter = new Adapter(this, options);
   var handler = adapter.createHandler();
 
   if (handler) {
@@ -134,7 +132,7 @@ RemoteObjects.prototype.connect = function(url, name) {
   }
 
   var Adapter = this.adapter(name);
-  var adapter = new Adapter(this, this.options);
+  var adapter = new Adapter(this);
   this.serverAdapter = adapter;
   return adapter.connect(urlWithoutAuth);
 };
--- a/lib/remote-objects.js
+++ b/lib/remote-objects.js
@@ -93,7 +93,9 @@ RemoteObjects.extend = function(exports) {
 
 RemoteObjects.prototype.handler = function(name, options) {
   var Adapter = this.adapter(name);
-  var adapter = new Adapter(this, options);
+  // send the set options for this remote object and
+  // the dynamic options passed in to create the adapter
+  var adapter = new Adapter(this, Object.assign({}, this.options, options));
   var handler = adapter.createHandler();
 
   if (handler) {
@@ -132,7 +134,7 @@ RemoteObjects.prototype.connect = function(url, name) {
   }
 
   var Adapter = this.adapter(name);
-  var adapter = new Adapter(this);
+  var adapter = new Adapter(this, this.options);
   this.serverAdapter = adapter;
   return adapter.connect(urlWithoutAuth);
 };
--- a/lib/rest-adapter.js
+++ b/lib/rest-adapter.js
@@ -42,7 +42,6 @@ var urlencoded = bodyParser.urlencoded;
 
 function RestAdapter(remotes, options) {
   EventEmitter.call(this);
-
   this.remotes = remotes;
   this.Context = HttpContext;
   this.options = options || (remotes.options || {}).rest;
--- a/lib/rest-adapter.js
+++ b/lib/rest-adapter.js
@@ -118,8 +118,6 @@ RestAdapter.prototype._extractAuth = function(remotes, invokeOptions) {
   if (auth || !invokeOptions) return auth;
   if (this.options && this.options.passAccessToken && invokeOptions &&
     invokeOptions.accessToken) {
-    console.log('Got the this.options.passAccessToken: ',
-      this.options.passAccessToken, invokeOptions);
     auth = {
       accessToken: invokeOptions.accessToken
     };
@@ -142,7 +140,6 @@ RestAdapter.prototype.invoke = function(method, ctorArgs, args, callback) {
   }
   var remotes = this.remotes;
   var restMethod = this.getRestMethodByName(method);
-  console.log('the rest method is : ', Object.keys(restMethod));
   var invokeOptions = restMethod.getArgByName('options', args);
   var auth = this._extractAuth(remotes, invokeOptions);
 
@@ -682,10 +679,8 @@ function RestMethod(restClass, sharedMethod) {
  */
 RestMethod.prototype.getArgByName = function(argName, invokedArgs) {
   var argValue;
-  console.log('this.accepts: ', this.accepts);
   if (!this.accepts || !this.accepts.length) return undefined;
   this.accepts.some(function(argProperty, i) {
-    console.log('compare: ', argProperty.arg.toLowerCase() , argName.toLowerCase());
     if (argProperty.arg && argProperty.arg.toLowerCase() === argName.toLowerCase()) {
       argValue = invokedArgs[i];
       return true;
--- a/lib/rest-adapter.js
+++ b/lib/rest-adapter.js
@@ -111,15 +111,17 @@ RestAdapter.prototype.connect = function(url) {
  * @param args
  * @private
  */
-RestAdapter.prototype._extractAuth = function(remotes, args) {
+RestAdapter.prototype._extractAuth = function(remotes, invokeOptions) {
   var auth = remotes.auth;
-  // check for the loopback options to pass along.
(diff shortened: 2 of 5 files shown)


slnode:

slnode:

slnode:

lib/rest-adapter.js
```diff
@@ -105,6 +105,27 @@ RestAdapter.prototype.connect = function(url) {
...
+ * loopback options if available
+ * @param remotes
+ * @param args
+ */
+RestAdapter.prototype.extractAuth = function(remotes, args) {
+  var auth = remotes.auth;
+  // check for the loopback options to pass along.
+  // It may have the access token that can be used
+  if (!auth && args.length) {
+    var options = args[args.length - 1];
```
bajtos: This assumes that `args` do not contain the `callback` as the last argument, can we rely on this? I don't remember how exactly remote invocation works.

lib/rest-adapter.js
```diff
@@ -105,6 +105,27 @@ RestAdapter.prototype.connect = function(url) {
   this.connection = url;
 };
 
+/**
+ * Get the auth from the remotes or from the
+ * loopback options if available
+ * @param remotes
+ * @param args
+ */
+RestAdapter.prototype.extractAuth = function(remotes, args) {
```
bajtos: I think this method should be private - the name should start with underscore and jsdoc should include `@private` annotation.

lib/rest-adapter.js
```diff
@@ -105,6 +105,27 @@ RestAdapter.prototype.connect = function(url) {
...
+  var auth = remotes.auth;
+  // check for the loopback options to pass along.
+  // It may have the access token that can be used
+  if (!auth && args.length) {
```
bajtos: Let's simplify this method by using reverse logic please.

```js
if (auth || !args.length) return auth;
// process options ...
```

bajtos:
@slnode ok to test

Traksewt:
Thanks @bajtos, I was having trouble testing the REST in that unit test. As the fix was the setting auth for one loopback server hitting another loopback server, should I be creating an extra express/loopback (so 2 in total) in this unit test? I couldn't tell of how I could test the rest adapter setting the access token using the single server.

I also added a config to enable passing through the LB Access Token, otherwise, people may inadvertently send their loopback token to other 3rd party REST endpoints.

lib/remote-objects.js
```diff
@@ -93,7 +93,9 @@ RemoteObjects.extend = function(exports) {
 
 RemoteObjects.prototype.handler = function(name, options) {
   var Adapter = this.adapter(name);
-  var adapter = new Adapter(this, options);
+  // send the set options for this remote object and
+  // the dynamic options passed in to create the adapter
+  var adapter = new Adapter(this, Object.assign({}, this.options, options));
```
bajtos: This does not look right to me. Remoting options are different from transport/adapter-specific options. Why is this change needed? Can we revert it please?

lib/remote-objects.js
```diff
@@ -132,7 +134,7 @@ RemoteObjects.prototype.connect = function(url, name) {
   }
 
   var Adapter = this.adapter(name);
-  var adapter = new Adapter(this);
+  var adapter = new Adapter(this, this.options);
```
bajtos: As I mentioned above, `this.options` are usually different from adapter options. For example (and IIRC), the configuration of REST adapter is specified in `rest` property of remoting options.

I think we should add `options` argument to `connect`, to make this consistent with `handler` method.

lib/rest-adapter.js
```diff
@@ -105,6 +104,27 @@ RestAdapter.prototype.connect = function(url) {
...
 };
 
+/**
+ * Get the auth from the remotes or from the
+ * loopback options if available
+ * @param remotes
+ * @param args
+ * @private
+ */
+RestAdapter.prototype._extractAuth = function(remotes, args) {
```
bajtos: On the second thought, this implementation is very loopback-specific. We are trying to keep strong-remoting decoupled from loopback itself - this package should provide extension hooks allowing loopback to inject loopback-specific behaviour.

In this particular case, the function parsing the accessToken from the options argument should live in LoopBack. It can be provided to RestAdapter for example via the `options` object which contains other related configuration too.

lib/rest-adapter.js
```diff
@@ -105,6 +104,27 @@ RestAdapter.prototype.connect = function(url) {
...
+  var auth = remotes.auth;
+  // check for the loopback options to pass along.
+  // It may have the access token that can be used
+  if (auth || !args.length) return auth;
+  var options = args[args.length - 1];
```
bajtos: I am not sure if it's a good idea to always assume the last argument is `options`. Can we perhaps parse remoting metadata for the invoked method to verify that it has `options` argument at the expected position?

test/http-invocation.test.js
```diff
@@ -198,6 +198,20 @@ describe('HttpInvocation', function() {
...
+        {accessToken: {id: 'abc'}});
+      var expectedReq = { method: 'GET',
+        url: 'http://base/testModel/testMethod',
+        protocol: 'http:',
+        json: true,
+        headers: {
+          Authorization: 'abc'
+        }
+      };
+      expect(inv.createRequest()).to.eql(expectedReq);
```
bajtos: This assert is checking too much. If we change the underlying implementation in the future (e.g. by adding more fields to the returned request), then this test will fail despite the fact that the implementation may still be correct.

Please check only that request headers contain the expected authorization.

```js
expect(inv.createRequest().headers).to.have.property('Authorization', 'abc');
```

bajtos:
> I also added a config to enable passing through the LB Access Token, otherwise, people may inadvertently send their loopback token to other 3rd party REST endpoints.

👍

bajtos:
> As the fix was the setting auth for one loopback server hitting another loopback server, should I be creating an extra express/loopback (so 2 in total) in this unit test? I couldn't tell of how I could test the rest adapter setting the access token using the single server.

I think a single server is enough. You can create the access token via javascript API by calling something like `server.models.User.create` and `server.models.User.login`, and then pass this access token in the `options` argument of the client method (doing remote invocation).

lib/remote-objects.js
Traksewt: OK Bajtos. Thanks for the heads up. I got a neater solution using the actual Rest properties that doesn't need config code changes. the 'options.rest' config goes directly to the rest adapter already.

lib/remote-objects.js
Traksewt: OK I don't need this at all anymore, not even for the connect. As the Adapter gets it off the rest config.

test/http-invocation.test.js
Traksewt: Np. I was copying all the following methods, but this change is more stable.

lib/rest-adapter.js
Traksewt: The main difference for this one, is that it is executed per request. E.g. for access tokens it changes for each different user hitting the server, that is why it doesn't fit in the other options object that is used to initialise the adapter. 

I'll think about it to see if I can set this through loopback code. One option is making it like a basic function to get the auth, that I can override in some sort of boot script that iterates over all datasources and sets the more complex extractAuth programmatically on the adapter. Apart from this boot script solution feeling a bit hacky, I was also hoping that this might have been useful to developers that want loopback servers to hit other loopback servers securely (e.g. microservices). I was hoping it would have worked out of the box (after setting the config flag). Open to ideas here.

Traksewt: I did your solution below of using the meta to find the options. All it is looking for is an accessToken in an options object. I made a generic way to get an arg by name on RestMethod.

Traksewt:
@bajtos please check

lib/rest-adapter.js
Traksewt: Done, with a getArgByName in RestMethod.

bajtos:
@Traksewt thank you for the pull request, I'll take a look next week. I see that our CI builds are failing because one of our dependencies dropped support for Node.js 0.x, I think I'll need to fix our CI config before landing this patch, to get a green build first.

bajtos:
Here is a pull request that should fix our CI in the 2.x branch: https://github.com/strongloop/strong-remoting/pull/428

bajtos:
@Traksewt sorry for realizing this only now: the changes you are proposing are effectively adding a new feature. Per [our LTS rules](https://loopback.io/doc/en/contrib/Long-term-support.html), we are not adding any new features to the version lines that are in LTS or maintenance mode, and the version line 2.x is in LTS right now.

Could you please rework your patch on top of the `master` (version 3.x) branch? 

GitHub made it recently possible to change the target branch of a pull request: click in "edit" button on the right end of pull request title, and pick `master` from the branch drop-down. I haven't done this myself before, I think you may need to run `git rebase -i master` and `git push -f` afterwards to fix the commit history.

Traksewt:
There were too many conflicts so made a new pull #430

PR #433, opened 2017-12-06
Avoid setting Content-Type on 304

zbarbuto:
### Description

Fixes a bug where loopback tries to set headers that have already been sent.

Note that express will strip out the `Content-Type` header if the response is 304:

https://github.com/expressjs/express/blob/351396f971280ab79faddcf9782ea50f4e88358d/lib/response.js#L208-L214

Loopback will try to set `Content-Type` to a default (typically `application/json`) if the header does not exist

https://github.com/strongloop/strong-remoting/blob/2153d183be457e8d8f76ca08b79fedb87b387b15/lib/http-context.js#L573-L575

This can lead to a weird edge case when doing tasks like rendering ejs where the following occurs:

Application:
- Tell res to render a template (e.g. an ejs page) from a remote method

Express
- Set some headers including `Content-Type: text/html` since we’re rendering ejs
- Realise that the `statusCode` is `304` so strip out some irrelevant headers including `Content-Type`

Loopback:
- Check to see if we have a `Content-Type` set (which we now don’t because 304 stripped it out)
- Try to set default `Content-Type` to `application/json`

Express:

- Throws error as headers have already been send (unhandled promise rejection)

#### Related issues

Nil

### Checklist

### DIFF
--- a/test/rest.test.js
+++ b/test/rest.test.js
@@ -2376,6 +2376,24 @@ describe('strong-remoting-rest', function() {
       });
   });
 
+  it('does not default content-type to application/json if response is 304', () => {
+    const method = givenSharedStaticMethod(
+      cb => cb(null, {key: 'value'}),
+      {returns: {arg: 'result', type: 'object'}}
+    );
+    return request(app).get(method.url)
+      .expect(200)
+      .then(res => {
+        expect(res.get('Content-type')).to.exist();
+        return request(app).get(method.url)
+        .set('If-None-Match', res.get('etag'))
+        .expect(304);
+      })
+      .then(res => {
+        expect(res.get('Content-type')).to.not.exist();
+      });
+  });
+
   describe('client', function() {
     describe('call of constructor method', function() {
       it('should work', function(done) {
--- a/lib/http-context.js
+++ b/lib/http-context.js
@@ -570,7 +570,7 @@ HttpContext.prototype.done = function(cb) {
   }
   var dataExists = typeof data !== 'undefined';
   var operationResults = this.resolveReponseOperation(accepts);
-  if (!res.get('Content-Type')) {
+  if (res.stautsCode !== 304 && !res.get('Content-Type')) {
     res.header('Content-Type', operationResults.contentType);
   }
   if (dataExists) {


zbarbuto:
Please advise of failures as CI is closed to the public 😞

test/rest.test.js
```diff
@@ -2376,6 +2376,25 @@ describe('strong-remoting-rest', function() {
...
 
+  it('does not default content-type to application/json if response is 304', function(done) {
+    var method = givenSharedStaticMethod(
+      cb => cb(null, {key: 'value'}),
+      {returns: {arg: 'result', type: 'object'}}
+    );
+    request(app).get(method.url)
+      .expect(200)
+      .end((err, res) => {
+        expect(res.get('Content-type')).to.exist();
```
bajtos:
```
if (err) return done(err);
```

test/rest.test.js
```diff
@@ -2376,6 +2376,25 @@ describe('strong-remoting-rest', function() {
...
+        request(app).get(method.url)
+        .set('If-None-Match', res.get('etag'))
+          .expect(304)
+          .end((err, res) => {
+            expect(res.get('Content-type')).to.not.exist();
```
bajtos:
```
if (err) return done(err);
```

bajtos:
> Please advise of failures as CI is closed to the public

Please take a look at Travis CI, that on is publicly visible.

```
 1) strong-remoting-jsonrpc handlers jsonrpc should support calling object methods:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

The lack of visibility into cis-jenkins build failures is a know pain point :(

zbarbuto:
@bajtos Thanks for feedback. Have switched to promise-based test and switched from `var` to `const`.

bajtos:
> [cis-jenkins] x64 && linux && nvm,4 — Failed! (34971e5)

The failure is unrelated to the changes made in this pull request, our CI was not able to fetch some dependencies.

bajtos:
Landed, thank you @zbarbuto for the contribution!

PR #437, opened 2017-12-19
Update sse to 0.0.7

smartmouse:
#### Related issues

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "Remoting",
     "REST"
   ],
-  "version": "2.34.0",
+  "version": "2.34.1",
   "publishConfig": {
     "tag": "lts"
   },
@@ -35,7 +35,7 @@
     "mux-demux": "^3.7.9",
     "qs": "^6.2.1",
     "request": "^2.55.0",
-    "sse": "0.0.6",
+    "sse": "0.0.7",
     "strong-globalize": "^2.6.6",
     "traverse": "^0.6.6",
     "xml2js": "^0.4.8"


package.json
```diff
@@ -7,7 +7,7 @@
     "Remoting",
     "REST"
   ],
-  "version": "2.34.0",
+  "version": "2.34.1",
```
rmg: The version will automatically be updated when the release is cut. Incrementing it here means that v2.34.1 will be skipped unless it is specified when running `slt-release`.

Note that this is completely acceptable, I just wanted to make sure it was known.

smartmouse:
Do you know why CI for nvm 4 fails?  node 6,8 passed though.

package.json
smartmouse: yes, good to know that the version is being handled automatically because it will be confusing if we merged some new changes and the version stays  the same as before.

rmg:
CI is failing on node v4 because npm v2 can't install the dependencies properly - because the phantomjs-prebuilt uses dependencies during its install script but doesn't bundle them to ensure they are present. It's a bug in phantomjs-prebuilt that is unlikely to be fixed (I've tried before, nobody cares enough).

smartmouse:
In ourJenkins build, can we use npm 3.10.10 with node 4.x? I think we should up our minimum npm version to npm 3.x.
Is it save to merge ? in our staging CI build , it is using npm 3.

package.json
bajtos: @smartmouse it's actually our convention in LoopBack to change the `version` field only when publishing a new release. Not a big deal here, but please don't update `version` field in the future.

PR #438, opened 2018-01-24
Fix typo in status code check

zbarbuto:
### Description

Oops, there was a typo in my prior pull request #433

### DIFF
--- a/test/rest.test.js
+++ b/test/rest.test.js
@@ -64,6 +64,14 @@ describe('strong-remoting-rest', function() {
     objects.connect('http://localhost:' + server.address().port, adapterName);
   });
 
+  before(() => {
+    process.on('unhandledRejection', unhandledRejection);
+  });
+
+  after(() => {
+    process.removeListener('unhandledRejection', unhandledRejection);
+  });
+
   function json(method, url) {
     if (url === undefined) {
       url = method;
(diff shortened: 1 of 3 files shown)


bajtos:
Nice! Please add a test to cover this change and prevent regressions in the future. The test should be failing on the current master and pass with your typo fixes.

I think you need to write a method that explicitly sets response `Content-Type` and returns a non-304 status code?

The tests are failing on CI because of a linter error, could you please fix it as part of your pull request? See https://travis-ci.org/strongloop/strong-remoting/jobs/332629527#L1533

```
/home/travis/build/strongloop/strong-remoting/lib/remote-objects.js
  192:1  error  Expected indentation of 4 spaces but found 6  indent
✖ 1 problem (1 error, 0 warnings)
  1 error, 0 warnings potentially fixable with the `--fix` option.
```

test/rest.test.js
```diff
@@ -3040,4 +3064,8 @@ describe('strong-remoting-rest', function() {
 
     return sharedClass.defineMethod(methodName, options);
   }
+
+  function unhandledRejection(err) {
+    throw new Error('Unhandled promise rejection found in test', err);
```
bajtos: This produces error message that's not very helpful:

```
   Uncaught Error: Unhandled promise rejection found in test
      at process.unhandledRejection (test/rest.test.js:3069:11)
      at processEmit [as emit] (node_modules/nyc/node_modules/signal-exit/index.js:155:32)
      at emitPendingUnhandledRejections (internal/process/promises.js:94:22)
      at process._tickCallback (internal/process/next_tick.js:189:7)
```

Let's throw the original error instead, here is what I am seeing with the change in place:

```
    Uncaught Error: Can't set headers after they are sent.
      at validateHeader (_http_outgoing.js:494:11)
      at ServerResponse.setHeader (_http_outgoing.js:501:3)
      at ServerResponse.header (node_modules/express/lib/response.js:767:10)
      at HttpContext.done (lib/http-context.js:84:475)
      at lib/rest-adapter.js:84:1395
      at node_modules/async/dist/async.js:3874:9
      [etc.]
```

bajtos:
Landed 🎉  Thank you for the contribution! ❤️

PR #439, opened 2018-01-29
chore: upgrade dependencies

raymondfeng:
Upgrade EventEmitter2 to latest version to better align with
Node.js EventEmitter APIs. We need this to remove remote hooks.

#### Related issues

- connect to <link_to_referenced_issue>

### Checklist


zbarbuto:
Perhaps an item for discussion but what are thoughts on using something like [greenkepper](https://greenkeeper.io/) or [renovate](https://www.npmjs.com/package/renovate) to keep an eye on dependency updates and automate PRs for them?

bajtos:
> thoughts on using something like greenkepper or renovate to keep an eye on dependency updates and automate PRs for them?

The last time I was looking into this, I found out that such tools don't work well for us:

 - We are automatically picking all backwards-compatible releases (semver-minor, semver-patch) thanks to `^` operator used in version specs.
 - As a result, we need automated pull requests only for semver-major releases. At that time, I did not find a way how to configure dependency updaters to ignore minor and patch releases.

However, the landscape may have changed since then. @zbarbuto if you have bandwidth to look into this, then please go ahead!

PR #446, opened 2018-05-31
Upgrade strong-globalize + drop Node.js 4.x + update deps

bajtos:
Drop support for Node 4.x
 - Needed by the new smaller strong-globalize. Since Node.js 4.x is no longer supported by Node.js project, it's ok to drop it in semver-minor version.
- Travis: add Node.js 10.x to the build matrix 
- Upgrade mocha to latest 
- Upgrade js2xmlparser to latest 
- Upgrade chai & dirty-chai to latest 
- Upgrade remaining dependencies to latest versions
- Disable package-lock

As a nice side effect, this fixes a bunch of security warnings reported by `npm audit`.

Connect to https://github.com/strongloop-internal/scrum-apex/issues/372


bajtos:
Hmm, we cannot land this until `loopback` drops support for Node.js 4.x from its CI build config.

My preliminary test shows that the new strong-globalize does not work in PhantomJS:

```
TypeError: Cannot call a class as a function
```

I suspect we will either need to fix in strong-globalize or else change Karma config in loopback to use headless Chrome (or Firefox) instead of PhantomJS. I can investigate next week.

bajtos:
@slnode test please

bajtos:
FWIW, loopback does not run CI tests on Node.js 4.x anymore, see https://github.com/strongloop/loopback/pull/3908

bajtos:
@slnode test please

PR #447, opened 2018-06-07
Upgrade dependencies to fix security vulnerabilities (2.x)

bajtos:
- Simplify build (backport #435):
   - Remove e2e tests that are no longer runnable anyways
   - Remove dependency on karma, phantomjs and browserify, they were needed for e2e tests only
   - Replace Grunt with npm scripts
- Upgrade jayson from 1.x to 2.x (backport  #442)
- Update deps to fix security vulnerabilities
- Travis: enable Node.js 10.x

This fix is needed to make `npm test` in loopback-workspace pass again. Right now, applications scaffolded for LB 2.x are failing their `npm test` because of security vulnerabilities reported by `nsp check`.


bajtos:
Please note that the latest versions of jscs & jshint are depending on a lodash version that's vulnerable to [Prototype Pollution](https://nodesecurity.io/advisories/577). Since these are dev-dependencies, I think we don't need to worry.

If we wanted to fix the problem, we would have to migrate from jshint+jscs to eslint+eslint-config-loopback.

PR #449, opened 2018-06-12
Update strong error handler

shimks:
### Description
related to https://github.com/strongloop/strong-error-handler/issues/75

#### Related issues

- connect to <link_to_referenced_issue>

### Checklist

- [ ] New tests added or existing tests modified to cover all changes
- [ ] Code conforms with the [style


package.json
```diff
@@ -18,41 +18,41 @@
     "node": ">=6"
   },
   "dependencies": {
-    "async": "^2.0.1",
-    "body-parser": "^1.12.4",
+    "async": "^2.6.1",
```
bajtos: When updating dependencies, we usually update only semver-major numbers and leave semver-minor/patch numbers untouched - the carrot operator `^` will take care of them. 

Please note that `npm update` does not update across semver-major numbers. If this commit is the outcome of `npm update`, then you need to check if there aren't any dependencies that need manual update.

Instructions:

1. Run `npm outdated --depth 0` to find which dependencies are outdated.
2. Run `npm install {dep}@latest` to update the dependency called `{dep}`.

test/rest.test.js
```diff
@@ -1578,7 +1578,6 @@ describe('strong-remoting-rest', function() {
           .end(function(err, res) {
             if (err) return done(err);
             var expectedDetail = res.body.error;
-            delete expectedDetail.statusCode;
```
jannyHou: hmm...any reason to remove the status code?

virkt25: don't think it should've been deleted to begin with ... tests all still pass. @shimks can confirm

shimks: There's a breaking change in strong-error-handler 3.0.0 where we _do_ include the status code for each individual errors in the array

dhmlau:
@slnode test please

PR #456, opened 2018-10-12
Update LTS status

bajtos:
`strong-remoting@3.x` is _Active LTS_ now, `strong-remoting@2.x` moved to _Maintenance LTS_.

See https://github.com/strongloop/loopback-next/issues/1802

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,18 @@
 # strong-remoting
 
+**This module is in Active LTS mode, new features are no longer accepted.**
+<br/>(See [Module Long Term Support Policy](#module-long-term-support-policy)
+below.)
+
+LoopBack 3 users looking for new features are encouraged to upgrade
+to LoopBack 4. Refer to
+[loopback-next#1849](https://github.com/strongloop/loopback-next/issues/1849)
+for more information on how to upgrade.
+
+If you are building directly on top of strong-remoting without LoopBack,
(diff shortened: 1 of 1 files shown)


hacksparrow:
👍

dhmlau:
@bajtos, just a question. I'm not familiar with what `strong-remoting` does, so I'm wondering how can someone upgrade to LoopBack 4.  Perhaps that might be something that can be included in the REAMDE?

bajtos:
> @bajtos, just a question. I'm not familiar with what strong-remoting does, so I'm wondering how can someone upgrade to LoopBack 4. Perhaps that might be something that can be included in the README?

Good point!

strong-remoting is sort of an equivalent of `@loopback/rest`. It implements a transport layer exposing LB3 Model classes via a network transport like REST, JSON RPC or Socket.io (websockets).

Upgrading an LB3 app to an LB4 app is something we need to figure out yet - see https://github.com/strongloop/loopback-next/issues/1849

Migrating an application built directly on top of strong-remoting (without loopback) to LB4 would be a lot of work because pretty much everything is different in LB4. To be honest, I don't think there are many (if any) users depending on strong-remoting directly. I feel that writing a migration guide for strong-remoting users would be likely a waste of our precious time now. Let's wait until there is strong user demand for that.

At the same time, I see how the messaging proposed in my pull request can be confusing to readers. Let me expand it a bit to make the options more clear.

bajtos:
@dhmlau LGTY now?

PR #470, opened 2019-10-03
Update eslint & config to latest

bajtos:
It's probably best to review the changes commit-by-commit:

- Update eslint & config to latest 42602e5
- fixup! `eslint . --fix` 6ffd98d
- fixup! fix remaining violations  fd7ab33


bajtos:
> Looks good. Any reason for changing `options` to `opts` ?

Good catch, I am glad you asked! In the original code, the following line is longer than 90 characters allowed by our eslint config:

```js
const sharedMethod = new SharedMethod(myFunction, 'myName', mockSharedClass, options);
```

The rename makes it short enough to pass the check :)

```js
const sharedMethod = new SharedMethod(myFunction, 'myName', mockSharedClass, opts);
```

PR #472, opened 2019-10-04
Fix handling of anonymous object types

bajtos:
In LoopBack, the method `PersistedModel.updateAll` describes the return type using an inline (anonymous) definition:

```
{
  returns: {
    arg: 'info',
    type: {
      count: {
        type: 'number',
        description: 'The number of instances updated',
      },
    },
    root: true,
  },
}
```

Before this change, remote invocation of updateAll was crashing the process with the following error:

```
AssertionError: type must be either an array or a string.
```

This patch fixes the problem as follows:

- A try/catch block is introduced to correctly report any sync errors via the callback

- The type registry was improved to recognize anonymous object types and use the "object" converter for them.

#### Related issues

Fixes https://github.com/strongloop/loopback/issues/3717

/cc @angfal @ewrayjohnson

### Checklist

- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)

### DIFF
--- a/test/rest.browser.test.js
+++ b/test/rest.browser.test.js
@@ -29,7 +29,9 @@ describe('strong-remoting-rest', function() {
 
   // setup
   beforeEach(function() {
-    objects = RemoteObjects.create();
+    objects = RemoteObjects.create({
+      errorHandler: {debug: true},
+    });
     remotes = objects.exports;
 
     // connect to the app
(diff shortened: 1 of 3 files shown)


test/rest.browser.test.js
```diff
@@ -29,7 +29,10 @@ describe('strong-remoting-rest', function() {
 
   // setup
   beforeEach(function() {
-    objects = RemoteObjects.create();
+    objects = RemoteObjects.create({
+      handleErrors: false,
```
jannyHou: hmm, why `handleError` is set to false?

bajtos: Good catch. I was experimenting with different error-handler settings to get a more helpful test failure message. I think this line is not needed, I'll remove it.

Issue #482, opened 2020-03-23
Server stream not closed if requested as application/json 


Wizarth:
I've provided a unit test at: https://github.com/Wizarth/strong-remoting/tree/json-zombie-streams

## Steps to reproduce

Run the unit test in stream.js: 
> a function returning a ReadableStream  
>> an http client requesting a stream as JSON  
>>> should close server stream on client disconnect

This case is caused by HttpInvocation.prototype.createRequest setting json: true , which is then used in HttpInvocation.prototype.invoke , when methodReturnsStream is true.

Previously reported as Loopback issue https://github.com/strongloop/loopback/issues/4317 , which includes use of loopback-sandbox to replicate the issue.

## Steps to reproduce

Using the remote, connecting to change-stream creates an observer on the model that is never removed.

With this sandbox:
In separate shells

    NODE_ENV=backend node server/server.js
    NODE_ENV=frontend node server/server.js
    node test.js

## Current Behavior

backend is on localhost:3000
frontend is on localhost:3010

backend is using in-memory db for Test model.
frontend is using loopback-connector-remote, connecting to backend, for Test model.

server.js has a once per second check for the # of _observers['after save'] on the Test model.

test.js runs two tests. The first connects 100 times to http://localhost:3000/api/Tests/change-stream (using request). After ten seconds, these requests are aborted.

During this time, the backend loopback instance reports 100 after save observers. When the requests are disconnected, the number of observers drops to zero.

In the second test, 100 connections to http://localhost:3010/api/Tests/change-stream are created (the frontend). After ten seconds, these requests are aborted.

During this time, the backend loopback instance reports 100 after save observers. When the requests are disconnected, the number of observers does not drop.

Terminating the frontend instance does not cause the number of observers to drop.

## Expected Behavior

Disconnecting the HTTP request should remove the observer.

## Link to reproduction sandbox

https://github.com/Wizarth/loopback-sandbox/tree/change-stream_Observer

## Additional information

Node version: 

- win32 x64 10.16.0
- linux x64 8.14.0

Sandbox:
loopback-sandbox@1.0.0 C:\...\loopback-sandbox
+-- loopback@3.26.0
+-- loopback-boot@3.3.1
+-- loopback-component-explorer@6.4.0
+-- loopback-connector-remote@3.4.1
+-- loopback-connector-rest@3.6.0

[...]

PR #483, opened 2020-04-01
Json zombie streams

Wizarth:
Fixes https://github.com/strongloop/strong-remoting/issues/482
Includes test case.

### DIFF
--- a/test/streams.js
+++ b/test/streams.js
@@ -13,6 +13,7 @@ const express = require('express');
 const request = require('supertest');
 const fs = require('fs');
 const es = require('event-stream');
+const http = require('http');
 const EventSource = require('eventsource');
 
 describe('strong-remoting', function() {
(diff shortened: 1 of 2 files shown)


slnode:

bajtos:
Thank you @Wizarth for the pull request, I appreciate the test case! I'll try to review the changes in the next few (working) days.

bajtos:
@slnode ok to test

test/streams.js
```diff
@@ -180,6 +180,27 @@ describe('a function returning a ReadableStream', function() {
...
+    it('should close server stream on client disconnect',
+      function(done) {
+        const r = request(app)
+          .get('/StreamClass/createInfiniteStream')
+          .set('Accept', 'application/json')
+          // .set('Accept', 'text/event-stream')
+          .timeout(100)
+          .buffer(false);
+        r
+          .then(function() {
```
bajtos, line 193: Mixing promises (`.then()` API) with callbacks (`done` argument) is usually a recipe for a disaster. I'll take a look on how to rewrite this test to be either fully callback-based or fully promise-based.

BTW when I run this new test agains the current master, the test fails with the following error:

```
 1) a function returning a ReadableStream
       an http client requesting a stream as JSON
         should close server stream on client disconnect:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/bajtos/src/strong/remoting/test/streams.js)
```

I am not sure if that's intended?

bajtos: I managed to find a better way how to test this, will push the changes to your feature branch soon.

bajtos:
@Wizarth Thank you again for the pull request. I added a new commit f7c066c where I have rewritten your test to use Node.js core API `http.request`, which gives us more control. Now the test returns a descriptive error message when the server event stream was not closed within 200ms (which is configurable).

I was not able to get rid of the mixed promise/callback style because of the way how `streamClosed` was created. We could rewrite this part from Promises to a regular EventEmitter, but I feel it's not worth the effort now.

It allowed me to remove your 50ms timer for calling `req.abort()`, now we abort the request just after we receive the response from the server. On my machine, the new test completes in ~40ms now.

On the other hand, the new test code is a bit longer than your original version. I think the benefits of faster execution and more robust implementation are worth it.

What do you think?

If you are fine with my changes, then we can proceed to squash the commits and land the pull request. Let me know!

bajtos:
Also added 4986c52 to fix the new test on Node.js 8.x

Wizarth:
Looks good to me.

I based the test on the existing test using streamClosed, which looks like it would have used the timeout as failure.

Using the http client response event is much better than a timer, I've never used that module directly.

bajtos:
The fix is available in `strong-remoting@3.16.2`, thank you for the contribution! ❤️

